PR #1, opened 2023-10-13
Continuous deployment with Github Actions

BakerWJ:
### Description

Creates a cron every 2 hours that checks for new commits and deploys to Render if a new commit is found.

### Reason for Change

CI/CD is cool

### Type of change

- [ ] CI/CD

### Testing

Test in prod


ganeshasapu (approved):
lgtm

PR #12, opened 2023-10-14
Update .env.example with DIRECT_URL

BakerWJ:
### Description

Adds `DIRECT_URL` variable to `.env.example`

### Reason for Change

Prisma commands will not work if this variable is not specified.

### Type of change


ganeshasapu (approved):
Do we acc need DIRECT_URL for it to run?

BakerWJ:
don't need DIRECT_URL to run, but I think you need it for db:push and db:migrate

PR #15, opened 2023-10-15
deleted health check

ganeshasapu:
### Description
Deleted unused health check file

### Reason for Change
File is no longer being used


BakerWJ (approved):
rip render

PR #16, opened 2023-10-15
Invalidate query cache on Mutation

BakerWJ:
### Description

Invalidates all queries when mutation is performed so that developers do not need to manually invalidate context. Also removes SSR because I don't want to debug that.

### Reason for Change

Setup for dev starter tasks

### Type of change


ganeshasapu (approved):
lgtm

PR #19, opened 2023-10-15
Finished CFD-16

sarinali:
### Description
Created a sarina component, added endpoint to router, added functionality to database to increase number of upvotes

### Reason for Change
CFD-16

### DIFF
--- a/apps/nextjs/src/components/developers/sarinali.tsx
+++ b/apps/nextjs/src/components/developers/sarinali.tsx
@@ -1,0 +1,33 @@
+// sarina component
+import { api } from "~/utils/api";
+
+const SarinaLi = () => {
+  const sarinaliInfo = api.developer.sarinali_info.useQuery().data;
+  const mutation = api.developer.sarinali_upvote.useMutation();
+  return (
+    <div>
+      <h1>sarinaliInfo from API:</h1>
+      {sarinaliInfo ? (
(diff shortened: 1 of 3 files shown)


apps/nextjs/src/components/developers/sarinali.tsx
```diff
@@ -0,0 +1,34 @@
+// sarina component
+
+import { api } from "~/utils/api";
+
+const SarinaLi = () => {
+  const sarinaliInfo = api.developer.sarinali_info.useQuery().data;
+  const mutation = api.developer.sarinali_upvote.useMutation();
```
BakerWJ, line 7: Would recommend better naming in future, but don't need to change for this task

packages/api/src/router/developer.ts
```diff
@@ -37,4 +37,48 @@ export const developerRouter = createTRPCRouter({
       data: { upvotes: { increment: 1 } },
     });
   }),
+
+  sarinali_info: publicProcedure.query(async ({ ctx }) => {
+    let upvotes = 0;
+    // check if exist in table
```
BakerWJ, line 43: nit: I generally don't like comments in code, prefer good variable naming and simplifying code. The code here probably doesn't need comments since you have good variable names

BakerWJ (approved):

PR #22, opened 2023-10-19
Add: comments model

jasonfyw:
### Description
Adds `Comments` model with specified fields and relation fields defining a one-to-many relationship between Posts and Comments.

### Reason for Change

Completes CFD-26

### Type of change

### Testing
Made example rows through Prisma Studio:
image

image


packages/db/prisma/schema.prisma
```diff
@@ -17,8 +17,17 @@ model Developers {
...
 model Posts {
-  id        Int      @id @default(autoincrement())
+  id        Int        @id @default(autoincrement())
   title     String
   contents  String
+  createdAt DateTime   @default(now())
+  comments  Comments[]
+}
+
+model Comments {
```
BakerWJ, line 27: Can we rename these to be singular (my mistake in the tickets)? so Post and Comment -- I think that will make more sense when you have something like `Comment[]`. Also makes more sense that a comment has only one post, but right now it is type `Posts` which is confusing.

BakerWJ (changes requested):

packages/db/prisma/schema.prisma
jasonfyw: Very fair point

BakerWJ (approved):

PR #31, opened 2023-10-21
Enable Automerge for Minor Releases

BakerWJ:
### Description

Prevents Renovate from automerging any major package changes

### Reason for Change

Let's make sure that our repo stays safe!

### Type of change

- [x] CI


ganeshasapu (approved):
bumping those commit numbers

PR #66, opened 2023-10-22
Sarinali/cfd 30 posts crud endpoints

sarinali:
### Description
Created CRUD endpoints for update delete create, updated my own component to remove a header.

### Reason for Change
https://linear.app/uoftblueprint/issue/CFD-30/posts-crud-endpoints 

Completes CFD-30

### DIFF
--- a/packages/api/src/router/post.ts
+++ b/packages/api/src/router/post.ts
@@ -1,4 +1,5 @@
 import { z } from "zod";
+
 import { createTRPCRouter, protectedProcedure } from "../trpc";
 
 export const postRouter = createTRPCRouter({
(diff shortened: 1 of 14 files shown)


ganeshasapu (approved):
lgtm

packages/api/src/router/post.ts
```diff
@@ -6,4 +8,42 @@ export const postRouter = createTRPCRouter({
       orderBy: { createdAt: "desc" },
     });
   }),
+  createPost: protectedProcedure
+    .input(z.object({ title: z.string(), contents: z.string() }))
```
BakerWJ, line 12: I think that both the title and contents should be non-empty. Might also want to set a maximum length for the title

packages/api/src/router/post.ts
```diff
@@ -6,4 +8,42 @@ export const postRouter = createTRPCRouter({
...
+        data: {
+          title: input.title,
+          contents: input.contents,
+        },
+      });
+    }),
+  updatePostByID: protectedProcedure
+    .input(
+      z.object({
+        id: z.number(),
```
BakerWJ, line 24: can you specify that the number needs to be >= 0

packages/api/src/router/post.ts
```diff
@@ -6,4 +8,42 @@ export const postRouter = createTRPCRouter({
...
+    )
+    .mutation(async ({ ctx, input }) => {
+      await ctx.db.post.update({
+        where: {
+          id: input.id,
+        },
+        data: {
+          title: input.title,
+          contents: input.contents,
+        },
```
BakerWJ, line 37: did you check that this doesn't change the value if the value isn't provided in the update?

BakerWJ (changes requested):
looks good -- just some small things that can be improved

packages/api/src/router/post.ts
sarinali: I wanted to clarify, does that mean whitespaces should be invalid? ie if i just make the title a space it should not allow me to execute the endpoint right?

packages/api/src/router/post.ts
sarinali: Yes, fields are not updated if input is undefined/left empty for one

packages/api/src/router/post.ts
sarinali: Ok!

packages/api/src/router/post.ts
BakerWJ: If you do `z.string().trim().min.....` it will remove the whitespace from the string before checking the length

packages/api/src/router/post.ts
```diff
@@ -6,4 +8,47 @@ export const postRouter = createTRPCRouter({
...
+      });
+    }),
+  deletePostByID: protectedProcedure
+    .input(z.object({ id: z.number().min(0) }))
```
BakerWJ, line 46: you can also use `z.number().nonnegative()`, might make more sense as a reader

packages/api/src/router/post.ts
```diff
@@ -6,4 +8,47 @@ export const postRouter = createTRPCRouter({
...
+        data: {
+          title: input.title,
+          contents: input.contents,
+        },
+      });
+    }),
+  updatePostByID: protectedProcedure
+    .input(
+      z.object({
+        id: z.number().min(0),
```
BakerWJ, line 29: can also use nonnegative here

packages/api/src/router/post.ts
sarinali: I just updated the branch, lmk what u think

BakerWJ: yep looks good

BakerWJ (approved):
other comments about id are nits, just choose whichever option you think makes the most sense. thanks for fast pr updates!

BakerWJ:
also make better pr titles 😠

packages/api/src/router/post.ts
sarinali: just fixed it!

packages/api/src/router/post.ts
sarinali: OK!

BakerWJ (approved):

ganeshasapu (approved):
lgtm

PR #94, opened 2023-10-28
[CFD-32] Add Activity Model

helenaglow:
### Description
Created an Activity Model in `schema.prisma` with:
- [X] `name`: `String` (as primary key)
- [X] `day`: `DateTime @db.Date` (_only shows date portion of timestamp_)
- [X] `startTime` : `DateTime @db.Time()` (_only shows time_)
- [X] `endTime` : `DateTime @db.Time()` (_only shows time_)
- [X] `leader` : `String`
- [X] `location` : `String`

(I have not migrated it yet as it says to only do it once per PR, so I will wait for approval first)

Completes CFD-32
Contributes to CFD-28

### Type of change

### Testing

The video shows changing the time on the `day` time stamp and it reverting as it is not needed. Same goes for `startTime` and `endTime`. 

https://github.com/uoftblueprint/centre-for-dreams/assets/113125436/5b862043-103d-4b21-9545-ab639f424e5a

### DIFF
--- a/packages/db/prisma/schema.prisma
+++ b/packages/db/prisma/schema.prisma
@@ -35,8 +35,8 @@ model Comment {
 model Activity {
   name      String   @id
   day       DateTime @db.Date
-  startTime DateTime @db.Timetz()
-  endTime   DateTime @db.Timetz()
+  startTime DateTime @db.Time()
+  endTime   DateTime @db.Time()
   leader    String
   location  String
 }
--- a/packages/db/prisma/schema.prisma
+++ b/packages/db/prisma/schema.prisma
@@ -33,7 +33,8 @@ model Comment {
 }
 
 model Activity {
-  name      String   @id
+  id        Int      @id
+  name      String
   day       DateTime @db.Date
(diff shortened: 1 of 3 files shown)


helenaglow:
I added `id` so that it is easier to change the activity name.

BakerWJ (approved):

packages/db/prisma/schema.prisma
```diff
@@ -31,3 +31,13 @@ model Comment {
   text      String
   createdAt DateTime @default(now())
 }
+
+model Activity {
+  id        Int      @id
```
BakerWJ, line 36: add autoincrement to id

BakerWJ (changes requested):
rip sorry just realized that `id` didn't have a default value. will approve after small fix

helenaglow:
Ahh I forgot that oops, thank you for catching 😅

BakerWJ:
Can you make a migration? Will approve once migration is created

BakerWJ (approved):

PR #96, opened 2023-10-28
[CFD-37] Create Absence CRUD Endpoints

1zhaohel:
### Description

Created a new tRPC router with endpoints for Absence.

- [x] createAbsence
- [x] getAllAbsences
- [x] getAbsenceByUserId
- [x] deleteAbsence

### Reason for Change

Completes CFD-37

Contributes to CFD-27

### Type of change

### DIFF
--- a/packages/api/src/router/absence.ts
+++ b/packages/api/src/router/absence.ts
@@ -23,7 +23,7 @@ export const absenceRouter = createTRPCRouter({
     return await ctx.db.absence.findMany();
   }),
 
-  getAbsenceByUserId: adminProcedure
+  getAbsences: protectedProcedure
     .input(
       z.object({
         userId: z.number().nonnegative(),
--- a/packages/api/src/router/absence.ts
+++ b/packages/api/src/router/absence.ts
@@ -1,0 +1,59 @@
+import { z } from "zod";
+
+import { createTRPCRouter, protectedProcedure } from "../trpc";
+
+export const absenceRouter = createTRPCRouter({
+  createAbsence: protectedProcedure
+    .input(
+      z.object({
+        userId: z.number().nonnegative(),
+        absenceDate: z.date(),
+      }),
+    )
+    .mutation(async ({ input, ctx }) => {
+      await ctx.db.absence.create({
+        data: {
+          userId: input.userId,
+          absenceDate: input.absenceDate,
+        },
+      });
+    }),
+
+  getAllAbsences: protectedProcedure.query(async ({ ctx }) => {
+    return await ctx.db.absence.findMany();
+  }),
+
+  updateAbsence: protectedProcedure
+    .input(
+      z.object({
+        id: z.number().nonnegative(),
+        userId: z.number().nonnegative().optional(),
+        absenceDate: z.date().optional(),
+      }),
+    )
+    .mutation(async ({ input, ctx }) => {
+      await ctx.db.absence.update({
+        where: {
+          id: input.id,
+        },
+        data: {
+          userId: input.userId,
+          absenceDate: input.absenceDate,
+        },
+      });
+    }),
+
+  deleteAbsence: protectedProcedure
+    .input(
+      z.object({
+        id: z.number().nonnegative(),
+      }),
+    )
+    .mutation(async ({ input, ctx }) => {
+      await ctx.db.absence.delete({
+        where: {
+          id: input.id,
+        },
+      });
+    }),
+});
--- a/packages/api/src/router/absence.ts
+++ b/packages/api/src/router/absence.ts
@@ -1,6 +1,6 @@
 import { z } from "zod";
 
-import { createTRPCRouter, protectedProcedure } from "../trpc";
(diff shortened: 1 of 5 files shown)


BakerWJ (changes requested):
looks good -- can you add this router to `root.ts` so that it can be used on the frontend + tested

packages/api/src/router/absence.ts
```diff
@@ -0,0 +1,59 @@
...
+        data: {
+          userId: input.userId,
+          absenceDate: input.absenceDate,
+        },
+      });
+    }),
+
+  getAllAbsences: protectedProcedure.query(async ({ ctx }) => {
+    return await ctx.db.absence.findMany();
+  }),
```
BakerWJ, line 24: can you convert this to an adminProcedure, then create another getAbsences that uses a userId?

packages/api/src/router/absence.ts
```diff
@@ -0,0 +1,59 @@
...
+    .mutation(async ({ input, ctx }) => {
+      await ctx.db.absence.update({
+        where: {
+          id: input.id,
+        },
+        data: {
+          userId: input.userId,
+          absenceDate: input.absenceDate,
+        },
+      });
```
BakerWJ, line 43: I don't think that this endpoint is necessary, since a person is either absent or present, so we only need create + delete

packages/api/src/router/absence.ts
```diff
@@ -0,0 +1,53 @@
...
+      z.object({
+        userId: z.number().nonnegative(),
+      }),
+    )
+    .query(async ({ input, ctx }) => {
+      return await ctx.db.absence.findMany({
+        where: {
+          userId: input.userId,
+        },
+      });
```
BakerWJ, line 37: sorry if previous comment was a bit unclear, but can you make this getAbsences. It would be a protetected procedure, and used by a parent to see all marked absences. It will take the user id for now, but in the future we will get  the user id  from our authentication

1zhaohel: no worries! thanks for the inputs, let me know if there's anything else to fix :)

packages/api/src/router/absence.ts
```diff
@@ -0,0 +1,53 @@
...
+      }),
+    )
+    .mutation(async ({ input, ctx }) => {
+      await ctx.db.absence.create({
+        data: {
+          userId: input.userId,
+          absenceDate: input.absenceDate,
+        },
+      });
+    }),
```
BakerWJ, line 20: do you know what happens if an absence already exists for this user + date combo?

BakerWJ:
rip kind of revisionist but I'm thinking that it might be better if we don't have an absenceId, and instead the model just has the userId and the date that they're missing. I don't think that the absenceId really makes sence, especially when I'm looking at these endpoints

1zhaohel:
> rip kind of revisionist but I'm thinking that it might be better if we don't have an absenceId, and instead the model just has the userId and the date that they're missing. I don't think that the absenceId really makes sence, especially when I'm looking at these endpoints

Okay, I'll remove the id from the absence db model.

packages/api/src/router/absence.ts
1zhaohel: no... will it create another absence for the user + date?

BakerWJ:
you can use `@@unique` for prisma to make it so that there can't be duplicate absences for each userId

packages/db/prisma/schema.prisma
```diff
@@ -43,7 +43,8 @@ model Activity {
 }
 
 model Absence {
-  id          Int      @id @default(autoincrement())
   userId      Int
-  absenceDate DateTime @db.Date
+  absenceDate DateTime
+
+  @@unique([absenceDate])
```
BakerWJ, line 49: this should also include the userId -- right now it just means that every absence date should be unique, but we want the user + date combo to be unique

packages/db/prisma/schema.prisma
```diff
@@ -43,7 +43,8 @@ model Activity {
 }
 
 model Absence {
-  id          Int      @id @default(autoincrement())
   userId      Int
-  absenceDate DateTime @db.Date
+  absenceDate DateTime
```
BakerWJ, line 47: is there a reason why you removed the `@db.Date`?

packages/api/src/router/absence.ts
BakerWJ: can you try it out and lmk what happens -- ideally, a new one should not be created

BakerWJ (commented):
looks good overall -- I'll approve once you resolve the remaining comments. Thanks for dealing with changing reqs for the ticket so quickly

packages/db/prisma/schema.prisma
1zhaohel: oh right tyty

packages/db/prisma/schema.prisma
1zhaohel: mb i thought it was the reason the check failed on one of the commits but that was just bc i didn't change the absence.ts file yet :o

BakerWJ (approved):

PR #98, opened 2023-10-29
[CFD 33 & 35] Activity CRUD & getSchedule Endpoints

helenaglow:
### Description

Added end points to:
- [X] ~`getActivity`: if input is a number then we get only one based on id, if its empty we get all~
- [X] `getSchedule`: Get weekly schedule based off of date 
- [X] `createActivity`: Create Activity
- [X] `updateActivity`: Update Activity
- [X] `deleteActivity`: Delete Activity
### Reason for Change
We need these endpoints!

Completes CFD-33 CFD-35

Contributes to CFD-28

### Type of change

### Testing

Tested with front end


helenaglow:
I will be at a halloween thing so I will fix whatever is necessary as soon as I can

packages/api/src/router/activity.ts
```diff
@@ -0,0 +1,95 @@
...
+    .mutation(async ({ ctx, input }) => {
+      await ctx.db.activity.create({
+        data: {
+          name: input.name,
+          day: new Date(input.day),
+          startTime: new Date("1970/01/01 " + input.startTime + "Z"),
+          endTime: new Date("1970/01/01 " + input.endTime + "Z"),
+          leader: input.leader,
+          location: input.location,
+        },
```
ganeshasapu, line 52: Should we be returning the new activity when an activity is successfully created?

packages/api/src/router/activity.ts
```diff
@@ -0,0 +1,95 @@
...
+        orderBy: { day: "desc" },
+      });
+    }),
+  createActivity: adminProcedure
+    .input(
+      z.object({
+        name: z.string(),
+        day: z.string(), // format YYYY-MM-DD
+        startTime: z.string(), // format HH:MM
+        endTime: z.string(), // format HH:MM
```
BakerWJ, line 38: lets use the proper zod validators here

packages/api/src/router/activity.ts
```diff
@@ -0,0 +1,95 @@
...
+        location: z.string(),
+      }),
+    )
+    .mutation(async ({ ctx, input }) => {
+      await ctx.db.activity.create({
+        data: {
+          name: input.name,
+          day: new Date(input.day),
+          startTime: new Date("1970/01/01 " + input.startTime + "Z"),
+          endTime: new Date("1970/01/01 " + input.endTime + "Z"),
```
BakerWJ, line 49: after proper validation we don't need this logic

packages/api/src/router/activity.ts
```diff
@@ -0,0 +1,95 @@
...
+      });
+    }),
+  updateAcitivty: adminProcedure
+    .input(
+      z.object({
+        id: z.number().nonnegative(),
+        name: z.string().optional(),
+        day: z.string().optional(), // format YYYY-MM-DD
+        startTime: z.string().optional(), // format HH:MM
+        endTime: z.string().optional(), // format HH:MM
```
BakerWJ, line 62: use proper validators here

packages/api/src/router/activity.ts
```diff
@@ -0,0 +1,95 @@
...
+  getSchedule: adminProcedure
+    .input(z.object({ day: z.string() }))
+    .query(async ({ ctx, input }) => {
+      return await ctx.db.activity.findMany({
+        where: {
+          day: {
+            gte: new Date(input.day),
+            lte: new Date(
+              new Date(input.day).getTime() + 7 * 24 * 60 * 60 * 1000,
+            ),
```
BakerWJ, line 26: I feel like there is a simpler way to do this

packages/api/src/router/activity.ts
```diff
@@ -0,0 +1,95 @@
...
+    .input(z.object({ id: z.number().nonnegative().optional() }))
+    .query(async ({ ctx, input }) => {
+      return input.id
+        ? await ctx.db.activity.findFirst({
+            where: { id: input.id },
+          })
+        : await ctx.db.activity.findMany({
+            orderBy: { day: "desc" },
+          });
+    }),
```
BakerWJ, line 16: You can delete this route -- we will only be dealing with the schedule itself

packages/api/src/router/activity.ts
```diff
@@ -0,0 +1,95 @@
...
+        },
+        data: {
+          name: input.name,
+          day: input.day ? new Date(input.day) : input.day,
+          startTime: input.startTime
+            ? new Date("1970/01/01 " + input.startTime + "Z")
+            : input.startTime,
+          endTime: input.endTime
+            ? new Date("1970/01/01 " + input.endTime + "Z")
+            : input.endTime,
```
BakerWJ, line 80: not a fan of this logic either -- feel like there is probably a simpler way to do it (hopefully)

packages/api/src/router/activity.ts
```diff
@@ -0,0 +1,95 @@
...
+        ? await ctx.db.activity.findFirst({
+            where: { id: input.id },
+          })
+        : await ctx.db.activity.findMany({
+            orderBy: { day: "desc" },
+          });
+    }),
+  getSchedule: adminProcedure
+    .input(z.object({ day: z.string() }))
+    .query(async ({ ctx, input }) => {
```
BakerWJ, line 19: this can be a `protectedProcedure` since the users of the app will also be able to see the schedule

packages/api/src/router/activity.ts
helenaglow: Fixed thank you for catching!

packages/api/src/router/activity.ts
helenaglow: I changed it to use date so we dont need the string validators anymore!

packages/api/src/router/activity.ts
helenaglow: Yea it is not pretty. It should be better now that I changed it to use z.date() instead of string

packages/api/src/router/activity.ts
helenaglow: I changed it to be I guess simpler on the eyes, but it's still is a bit clunky

helenaglow:
Fixed it to be z.date instead of string. I also tested it with using the front end. I did not get pictures because it was a scuffed way of testing

packages/api/src/router/activity.ts
```diff
@@ -0,0 +1,78 @@
...
+import { adminProcedure, createTRPCRouter, protectedProcedure } from "../trpc";
+
+export const activityRouter = createTRPCRouter({
+  getSchedule: protectedProcedure
+    .input(z.object({ day: z.date() }))
+    .query(async ({ ctx, input }) => {
+      return await ctx.db.activity.findMany({
+        where: {
+          day: {
+            gte: new Date(input.day),
```
BakerWJ, line 12: do you need to do new Date here?

BakerWJ (commented):
let me know if you want to make the changes with duration -- I can approve once that is done

packages/api/src/router/activity.ts
helenaglow: Yea...I am not sure why but it doesn't work if its not there...

helenaglow:
I think im gonna stick with end time. Just cause if we did duration then we would have to consider units... and I don't want to deal with that 🙈. Adding a constraint might be better.

BakerWJ:
I actually think that duration is a lot better -- would use minutes for duration

helenaglow:
Ok then I'll get on that

packages/db/prisma/schema.prisma
```diff
@@ -37,7 +37,7 @@ model Activity {
   name      String
   day       DateTime @db.Date
   startTime DateTime @db.Time()
-  endTime   DateTime @db.Time()
+  duration  Int
```
BakerWJ, line 40: can you make this name `durationMinutes` so that it is clear what the units are? make sure to delete the old db migration before creating a new one

helenaglow:
Ok, changed to duration instead of end time! Also caught some embarrassing typos

packages/api/src/router/activity.ts
```diff
@@ -0,0 +1,78 @@
...
+            lte: new Date(input.day.setDate(input.day.getDate() + 7)),
```
BakerWJ, line 13:
```suggestion
            gte: input.day,
            lte: new Date(input.day.getDate() + 7),
```

BakerWJ: this passed typechecks for me -- is there a reason why you said that it didn't work? (this is very last thing for PR then will approve 😰 )

helenaglow: Yea! It's really weird it doesn't have a typecheck error, but with testing it becomes incorrect. 
Screenshot 2023-10-29 at 2 35 58 PM


With `input.day`:
Screenshot 2023-10-29 at 2 29 26 PM

With `new Date(input.day)`:
Screenshot 2023-10-29 at 2 31 08 PM

Table for reference: 
Screenshot 2023-10-29 at 2 31 29 PM

helenaglow: Also, `new Date(input.day.getDate() + 7)` doesn't work either! Get date only returns the day of the month, so it wouldn't update the month or year. It's ugly but we'd either do it with set date or `new Date(input.day.getTime() + 7 * 24 * 60 * 60 * 1000)`

helenaglow: We could import `date-fns` or some other library that makes it easier to manipulate dates but idk.

BakerWJ: thanks for the good testing -- I think best option is to use date-fns here, since it will probably make the code more readable. You will just need to `pnpm add date-fns` in the `packages/api` folder to add it as a dependency

helenaglow: Omg it's so much ✨cleaner✨! It also fixed the weird issue with input.day for whatever reason

BakerWJ (approved):
great job -- thanks for being super responsive to all of the changes for the PR 😤

helenaglow:
NO! Thank you (🫵) for responding super fast and taking the time to review it so well! I'm glad that it's done 🥳

packages/api/src/router/activity.ts
BakerWJ: I think that `input.day` wasn't working originally because the value was being modified in the `setDate` call. `date-fns` creates a new date object for each of its functions, which is why using `input.day` works now

helenaglow: Ohhhhh yea that makes sense

PR #99, opened 2023-10-29
[CFD-39] Push notifications

jasonfyw:
### Description

Backend: creates endpoint to send a notification to a given list of devices

Frontend: adds methods to register the device for notifications, as well as an example component that can display incoming notifications and send a notification to self.

### Reason for Change

Completes CFD-39

### Type of change

### Testing

Using the push token displayed on the Expo app, notifications can successfully be sent when tested with the following methods:
- [Expo's push notification tool](https://expo.dev/notifications)
- tRPC panel
- The send notification button on the Expo app


BakerWJ:
hmmmm not sure why actions aren't being activated -- maybe try making an empty commit + pushing

apps/expo/package.json
```diff
@@ -51,7 +53,7 @@
     "eslint": "^8.52.0",
     "expo-doctor": "^1.1.3",
     "prettier": "^3.0.3",
-    "tailwindcss": "3.3.3",
+    "tailwindcss": "^3.3.2",
```
BakerWJ, line 56: why did this change happen?

packages/api/src/router/notification.ts
```diff
@@ -0,0 +1,54 @@
...
+          continue;
+        }
+        messages.push({
+          to: pushToken,
+          sound: "default",
+          title: input.title,
+          body: input.body,
+          subtitle: input.subtitle,
+        });
+      }
```
BakerWJ, line 37: can use a `map` here

packages/api/src/router/notification.ts
```diff
@@ -0,0 +1,54 @@
...
+  push: protectedProcedure
+    .input(
+      z.object({
+        pushTokens: z.array(z.string()),
+        title: z.string().optional(),
+        body: z.string().optional(),
+        subtitle: z.string().optional(),
+        // TODO: add optional data payload
+      }),
+    )
```
BakerWJ, line 19: this might be nice to keep for testing notifications in the future 🤔 -- I'm assuming that the blocker before integrating this with something like the newPost endpoint is that we don't have the user pushTokens stored in the db right now?

BakerWJ (commented):
will look in a bit more depth later -- we have a draft pr for implementing auth right now, so we might wait to merge this in until we have a user system set up (if that is one of the blockers for push tokens)

.vscode/settings.json
```diff
@@ -14,5 +14,7 @@
   "typescript.preferences.autoImportFileExcludePatterns": [
     "next/router.d.ts",
     "next/dist/client/router.d.ts"
-  ]
+  ],
+  "editor.detectIndentation": false,
+  "editor.tabSize": 2
```
BakerWJ, line 19: I think there is a way to make this your personal setting instead of editing the project `settings.json`

apps/expo/app.config.ts
```diff
@@ -28,11 +28,11 @@ const defineConfig = (): ExpoConfig => ({
...
   },
-  // extra: {
-  //   eas: {
-  //     projectId: "your-eas-project-id",
-  //   },
-  // },
+  extra: {
+    eas: {
+      projectId: "7712a44c-fe67-40d6-a8d4-428e2192af27",
+    },
```
BakerWJ, line 34: maybe this should be an environment variable? unfortunately we can't all join the same project for expo :(

apps/expo/src/app/notifications.tsx
BakerWJ, line 1: While good for testing, I don't think that we need this page on the mobile app (since most or all of our notifications are being sent from web)

apps/expo/src/notifications/registerNotifications.tsx
```diff
@@ -0,0 +1,39 @@
...
+  let token;
+
+  if (Platform.OS === "android") {
+    void Notifications.setNotificationChannelAsync("default", {
+      name: "default",
+      importance: Notifications.AndroidImportance.MAX,
+      vibrationPattern: [0, 250, 250, 250],
+      lightColor: "#FF231F7C",
+    });
+  }
```
BakerWJ, line 16: is there another check required for iOS, or is this only needed for android?

apps/expo/src/notifications/registerNotifications.tsx
```diff
@@ -0,0 +1,39 @@
...
+      await Notifications.requestPermissionsAsync();
+    }
+    const easConfig = Constants.expoConfig?.extra?.eas as { projectId: string };
+    if (easConfig) {
+      token = await Notifications.getExpoPushTokenAsync({
+        projectId: easConfig.projectId,
+      });
+    } else {
+      console.log("Unable to get projectId from Constants");
+    }
```
BakerWJ, line 31: maybe add this check to the start of the function (also probably want to throw an error)

apps/expo/src/notifications/registerNotifications.tsx
```diff
@@ -0,0 +1,39 @@
...
+    const { status: existingStatus } =
+      await Notifications.getPermissionsAsync();
+    if (existingStatus.valueOf() != "granted") {
+      await Notifications.requestPermissionsAsync();
+    }
+    const easConfig = Constants.expoConfig?.extra?.eas as { projectId: string };
+    if (easConfig) {
+      token = await Notifications.getExpoPushTokenAsync({
+        projectId: easConfig.projectId,
+      });
```
BakerWJ, line 28: you can just do a return here instead of having the `let token` at the top

packages/api/src/router/notification.ts
```diff
@@ -0,0 +1,47 @@
+import type { ExpoPushMessage } from "expo-server-sdk";
+import { Expo } from "expo-server-sdk";
+import { z } from "zod";
+
+import { createTRPCRouter, protectedProcedure } from "../trpc";
+
+const expo = new Expo({ accessToken: process.env.EXPO_ACCESS_TOKEN });
+
+export const notificationRouter = createTRPCRouter({
+  push: protectedProcedure
```
BakerWJ, line 10: this should probably be an `adminProcedure` if we keep it

BakerWJ (commented):
looks good overall -- just some minor changes + questions

.vscode/settings.json
jasonfyw: I use `tabSize` 4 and `detectIndentation` false for most other projects. Is there a way to modify workspace settings without changing `settings.json`?

apps/expo/src/app/notifications.tsx
jasonfyw: Ok, should I scrap this component then?

apps/expo/src/notifications/registerNotifications.tsx
jasonfyw: The Expo docs only showed a check for android. It works fine on iOS so I assume it's fine

.vscode/settings.json
BakerWJ: can this go in your user settings instead? the only rules that should be in the project `settings.json`should be project-specific (ie linting + workspace configuration)

apps/expo/src/app/notifications.tsx
BakerWJ: If the expo push notification tool has the same functionality, then I think we can delete it

.vscode/settings.json
jasonfyw: Fair enough. I'll take it out and just change the settings on my end

apps/expo/src/notifications/registerNotifications.tsx
BakerWJ, line 1: this should be a ts file

packages/db/package.json
```diff
@@ -18,7 +18,7 @@
   },
   "dependencies": {
     "@cfd/logger": "^0.1.0",
-    "@prisma/client": "^5.5.2"
+    "@prisma/client": "^5.4.2"
```
BakerWJ, line 21: can u change this version back

BakerWJ (commented):
good to merge after updating these 2 comments -- can you also add something to .env.example for the expo project id?

packages/api/src/router/notification.ts
```diff
@@ -0,0 +1,47 @@
...
+        }));
+
+      // Send the messages
+      const chunks = expo.chunkPushNotifications(messages);
+      const tickets = [];
+      await (async () => {
+        for (const chunk of chunks) {
+          try {
+            const ticketChunk = await expo.sendPushNotificationsAsync(chunk);
+            console.log(ticketChunk);
```
BakerWJ, line 39: lol not gonna block on this but in next pr use the logger instead of console

apps/expo/src/notifications/registerNotifications.ts
```diff
@@ -0,0 +1,35 @@
...
+    const { status: existingStatus } =
+      await Notifications.getPermissionsAsync();
+    if (existingStatus.valueOf() != "granted") {
+      await Notifications.requestPermissionsAsync();
+    }
+    return await Notifications.getExpoPushTokenAsync({
+      projectId: easConfig.projectId,
+    });
+  } else {
+    alert("Must use physical device for Push Notifications");
```
BakerWJ, line 31: i wonder if push notifications even work on simulator?

BakerWJ (dismissed):

packages/api/src/router/notification.ts
```diff
@@ -0,0 +1,46 @@
...
+      // Send the messages
+      const chunks = expo.chunkPushNotifications(messages);
+      const tickets = [];
+      await (async () => {
+        for (const chunk of chunks) {
+          try {
+            const ticketChunk = await expo.sendPushNotificationsAsync(chunk);
+            tickets.push(...ticketChunk);
+          } catch (error) {
+            console.error(error);
```
BakerWJ, line 41: `logger.error` 🥲

BakerWJ (approved):
would recommend turning on auto merge bc of ios build

packages/api/src/router/notification.ts
jasonfyw: Oh my bad lol I'll use the logger in my next pr

PR #102, opened 2023-10-29
Create user table

BakerWJ:
### Description

Creates a basic user table with only an ID

### Reason for Change

Needed for PRs that need users for relations.

### Type of change

### DIFF
--- a/packages/db/prisma/schema.prisma
+++ b/packages/db/prisma/schema.prisma
@@ -11,6 +11,11 @@ datasource db {
   directUrl = env("DIRECT_URL")
 }
 
+model User {
+  id       Int       @id @default(autoincrement())
+  absences Absence[]
+}
+
 model Developers {
   name    String @id
   upvotes Int    @default(0)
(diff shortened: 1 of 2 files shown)


packages/db/prisma/schema.prisma
```diff
@@ -11,6 +11,11 @@ datasource db {
   directUrl = env("DIRECT_URL")
 }
 
+model User {
+  id       String    @id
```
BakerWJ, line 15: using a string for id because Clerk IDs are strings, for example `user_2XRbvOxY8SRnRw5oGFlXcYwN9J8`

RyanL123:
As discussed offline, we'll use a separate table for users that separates `id` from `clerkId`. Reason is we should be thinking of our db as the place for metadata we define for the user (isAdmin, isApproved, etc), and clerk for user profile related metadata (userName, email, etc). This way we can also avoid unnecessary calls to Clerk for authorization related queries. Also, if we do decide to swap out clerk in the future for another authentication service this well nicely decouple our db from auth.

RyanL123 (approved):
LGTM, thanks for the update!

PR #124, opened 2023-11-04
[CFD-42] Add desktop login and signup page

RyanL123:
### Reason for Change

Completes CFD-42

Added login and signup pages as full redirects instead of modals.

### Type of change

### Testing

image
image


apps/nextjs/src/middleware.ts
```diff
@@ -1,11 +1,11 @@
...
-export default authMiddleware();
+export default authMiddleware({
+  publicRoutes: ["/"], // Do not protect '/' for now since that's where we land
+});
 
-// Do not protect '/' for now since that's where we land
+// See https://nextjs.org/docs/pages/building-your-application/routing/middleware#matcher for what config does
 export const config = {
-  matcher: ["/((?!.+\\.[\\w]+$|_next).*)", "/(api|trpc)(.*)"],
+  matcher: ["/((?!.+\\.[\\w]+$|_next).*)", "/", "/(api|trpc)(.*)"],
```
BakerWJ, line 10: We don't need to protect the API routes, as we are already running our own middleware on those routes. I think that you can just remove it from the matcher

apps/nextjs/src/pages/index.tsx
```diff
@@ -31,7 +31,7 @@ export default function Home() {
         <RamRaghavSharma />
         <MinhLe />
         <div>
-          {isSignedIn ? "" : <SignInButton mode="modal" />}
+          {isSignedIn ? "" : <SignInButton />}
```
BakerWJ, line 34: I'm assuming that the `SignInButton` uses the env variable to redirect user to the correct page?

BakerWJ (changes requested):
Looks good and worked for me -- can you also update the values in `env.mjs` to something like `z.string().includes('/signin')` so that dev / build will fail someone is using an outdated env variable?

apps/nextjs/src/pages/index.tsx
RyanL123: Yes

RyanL123:
> Looks good and worked for me -- can you also update the values in `env.mjs` to something like `z.string().includes('/signin')` so that dev / build will fail someone is using an outdated env variable?

Let's use `z.literal` then to be more explicit

apps/nextjs/src/middleware.ts
RyanL123: We should add it to the public routes then, since if we remove it altogether no TRPC request is able to authenticate due to the following error:

```
❌ tRPC failed on <no-path>: Clerk: getAuth() was called but Clerk can't detect usage of authMiddleware(). Please ensure the following:
- authMiddleware() is used in your Next.js Middleware.
- Your Middleware matcher is configured to match this route or page
```

After adding it to public routes all `publicProcedure` routes show up even when not signed in which is the intended behaviour.

BakerWJ (approved):
looks good :)

PR #126, opened 2023-11-05
[CFD-55] Add postType to post model

jasonfyw:
### Description

Adds a PostType enum (currently with two options: announcement and discussion) and adds a postType field to posts

### Reason for Change

Completes CFD-55

### Type of change

### Testing

New field and endpoint changes works as expected in Prisma Studio and tRPC panel

image

### DIFF
--- a/packages/api/src/router/post.ts
+++ b/packages/api/src/router/post.ts
@@ -1,6 +1,70 @@
+import type { SignedInAuthObject } from "@clerk/clerk-sdk-node";
+import type { PrismaClient } from "@prisma/client";
 import { z } from "zod";
 
-import { createTRPCRouter, protectedProcedure } from "../trpc";
+import { adminProcedure, createTRPCRouter, protectedProcedure } from "../trpc";
+
+const basePostInput = {
+  title: z.string().trim().min(1).max(300),
+  contents: z.string().min(1),
+};
+
+const createPostInput = z.object({
+  ...basePostInput,
+});
+
+const updatePostInput = z
+  .object({
+    id: z.number().nonnegative(),
+    ...basePostInput,
+  })
+  .partial();
+
+const deletePostInput = z.object({
+  id: z.number().nonnegative(),
+});
+
+const createPost = async (
+  ctx: { db: PrismaClient; auth: SignedInAuthObject },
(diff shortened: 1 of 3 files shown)


BakerWJ (approved):

packages/api/src/router/post.ts
```diff
@@ -13,13 +13,15 @@ export const postRouter = createTRPCRouter({
...
         contents: z.string().min(1),
+        postType: z.enum(["Announcement", "Discussion"]),
       }),
     )
     .mutation(async ({ ctx, input }) => {
       await ctx.db.post.create({
         data: {
           title: input.title,
           contents: input.contents,
+          postType: input.postType,
```
BakerWJ, line 24: can u not add this to the routes -- going to separate the routes for announcements and discussions (announcements will be admin only)

BakerWJ (changes requested):

packages/api/src/router/post.ts
BakerWJ: for now you can just use discussion as the default

jasonfyw: Made separate endpoints for discussions and announcements in d10b7b5. Let me know if that's what you meant

packages/api/src/router/post.ts
BakerWJ, line 1: can you take these changes out of this PR and potentially put them in a follow up? idk if there is much value in creating helper functions that take both Discussions + Announcements because they will actually have pretty different requirements (only notify on announcements, send emails on announcements, maybe some other stuff). For the scope in this PR, would prefer to just add the Discussion type to every post created in createPost

BakerWJ (changes requested):

packages/api/src/router/post.ts
```diff
@@ -39,6 +40,7 @@ export const postRouter = createTRPCRouter({
         data: {
           title: input.title,
           contents: input.contents,
+          postType: "Discussion",
```
BakerWJ, line 43: is this field optional? don't think you need to add it have it for the update operation?

jasonfyw: lol yeah true removed that line

BakerWJ (approved):

PR #138, opened 2023-11-08
Remove Client Logger Dependency

BakerWJ:
### Description

Removes `@cfd/logger` as a dependency of the Next.js app.

### Reason for Change

The logger was included for logging requests, but this should only be done on the server side, not on the client side. This also slightly reduces the bundle size of the web app.

### Type of change

### Testing

- [x] Ran Next Bundle Analyzer and confirmed that Pino is no longer included in the client bundle


ganeshasapu (approved):
w bundle size decrease

PR #140, opened 2023-11-08
Use Neon Serverless Driver with Prisma

BakerWJ:
### Description

Changed Prisma driver from the default Postgres driver to the Neon Serverless driver

### Reason for Change

Making DB connections is slow, contributing to poor cold starts on our Netlify deployment. The Neon driver uses HTTP requests instead of a DB connection, improving the performance of our DB queries

### Type of change

- [x] Perf (might break smth but it should be fine)

### Testing

I got the following results on my local machine, and repeated the experiments to ensure that results are consistent.

#### Cold starts
Default driver (before):

image

Serverless driver (after): 

image

#### Normal requests
Default driver (before): 

image

Serverless driver (after): 

image

### DIFF
--- a/packages/db/index.ts
+++ b/packages/db/index.ts
@@ -1,8 +1,18 @@
+import { neonConfig, Pool } from "@neondatabase/serverless";
+import { PrismaNeon } from "@prisma/adapter-neon";
 import { PrismaClient } from "@prisma/client";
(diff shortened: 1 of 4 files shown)


ganeshasapu (approved):
prisma neon adapter op

PR #144, opened 2023-11-09
[CFD-54] Create push tokens model

jasonfyw:
### Description

Adds a prisma model for push tokens for notifications. The model links a push token with a user.

### Reason for Change

Completes CFD-54

### Type of change

### Testing

The model works as expected when tested in Prisma Studio


apps/expo/src/app/index.tsx
```diff
@@ -5,6 +5,7 @@ import * as Notifications from "expo-notifications";
 import { Stack } from "expo-router";
 
 import registerForPushNotificationsAsync from "~/notifications/registerNotifications";
+import { logger } from "../../../../packages/logger";
```
BakerWJ, line 8: 😢 for now probably better to actually use console.log on frontend, logger is meant to be used by only api for right now

packages/db/prisma/schema.prisma
```diff
@@ -57,6 +58,15 @@ model Absence {
   @@unique([absenceDate, userId])
 }
 
+model PushToken {
+  token   String
```
BakerWJ, line 62: shouldn't token by itself be unique?

apps/expo/src/app/index.tsx
BakerWJ: actually might be able to get it working on FE, but probably best to just use console.log for rn

packages/db/prisma/schema.prisma
jasonfyw: that's actually very fair lol

apps/expo/src/app/index.tsx
BakerWJ: ok i will get this setup by sunday so that we can have nice logging on FE too, just wait a couple of days 🥲

packages/db/prisma/schema.prisma
```diff
@@ -57,6 +58,13 @@ model Absence {
   @@unique([absenceDate, userId])
 }
 
+model PushToken {
+  token   String  @id
```
BakerWJ, line 62: I think it makes more sense to mark this field as @unique -- they are essentially the same but logically I think unique makes more sense

BakerWJ (approved):
thanks for quick fixes :)

PR #145, opened 2023-11-10
[CFD-56] Add announcement router

jasonfyw:
### Description

Adds announcement router paralleling the post router's features. Also modified post router to only list/update/delete discussions. Adds explicit check that a discussion/announcement with the provided id exists, otherwise throw 404.

### Reason for Change

Completes CFD-56

### Type of change

### Testing

Endpoints and error checking working as expected after testing on tRPC panel


packages/api/src/router/announcement.ts
```diff
@@ -0,0 +1,86 @@
+import { TRPCError } from "@trpc/server";
+import { z } from "zod";
+
+import { adminProcedure, createTRPCRouter } from "../trpc";
+
+export const announcementRouter = createTRPCRouter({
+  getAnnouncements: adminProcedure.query(async ({ ctx }) => {
```
BakerWJ, line 7: this can be a `protectedProcedure`

packages/api/src/router/announcement.ts
```diff
@@ -0,0 +1,86 @@
...
+      await ctx.db.post.update({
+        where: {
+          id: input.id,
+          postType: "Announcement",
+        },
+        data: {
+          title: input.title,
+          contents: input.contents,
+        },
+      });
```
BakerWJ, line 61: instead of having 2 queries, you can instead handle the `RecordNotFound` exception that Prisma will throw if the record does not exist

packages/api/src/router/announcement.ts
```diff
@@ -0,0 +1,86 @@
...
+          code: "NOT_FOUND",
+          message: "Announcement with the given ID not found",
+        });
+      }
+
+      await ctx.db.post.delete({
+        where: {
+          id: input.id,
+        },
+      });
```
BakerWJ, line 84: same comment as update -- can handle the `RecordNotFound` instead of using 2 queries

packages/api/src/router/post.ts
BakerWJ, line 1: if you want to save us another PR, you can also rename these routes / router to be for discussions instead of posts

BakerWJ (changes requested):
looks good - just some minor fixes

BakerWJ (approved):

PR #153, opened 2023-11-12
[CFD-61] Endpoints to query users in our database

RyanL123:
### Reason for Change

Completes CFD-61

- Add clerkId to the users table, a unique field mapping users in our database to their clerkId
- When users are returned in the auth state, we also return their id within OUR database, which is a monotonically increasing integer
- If the user isn't present in the database, we insert them in there with their clerkId

### Type of change

### Testing

Updated upvote for Ryan Li card on the home page to use protected procedure, which should create the corresponding user in the database if it did not exist before when the button is used.

### DIFF
--- a/packages/api/src/root.ts
+++ b/packages/api/src/root.ts
@@ -4,6 +4,7 @@ import { announcementRouter } from "./router/announcement";
 import { developerRouter } from "./router/developer";
 import { discussionRouter } from "./router/discussion";
 import { notificationRouter } from "./router/notification";
+import { userRouter } from "./router/user";
 import { createTRPCRouter } from "./trpc";
 
 export const appRouter = createTRPCRouter({
@@ -13,6 +14,7 @@ export const appRouter = createTRPCRouter({
   activity: activityRouter,
   absence: absenceRouter,
   notification: notificationRouter,
+  user: userRouter,
 });
 
 // export type definition of API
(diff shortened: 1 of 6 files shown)


packages/api/src/trpc.ts
```diff
@@ -107,13 +115,27 @@ export const publicProcedure = t.procedure;
...
 
-const isAuthed = t.middleware(({ next, ctx }) => {
+const isAuthed = t.middleware(async ({ next, ctx }) => {
   if (!ctx.auth.userId) {
     throw new TRPCError({ code: "UNAUTHORIZED" });
   }
+
+  // If the user isn't in our database yet, insert them
+  const userId = ctx.auth.userId;
+  const user = await ctx.db.user.upsert({
```
RyanL123, line 125: Since we're already going to fetch the userId and append it to the context, we might as well do the insertion here as well if they don't exist

packages/api/src/router/user.ts
BakerWJ, line 1: should this be removed from the PR?

packages/api/src/router/developer.ts
BakerWJ, line 1: revert

packages/api/src/root.ts
BakerWJ, line 1: probably can be removed as well

BakerWJ (commented):
looks good -- will approve after comments are resolved

packages/api/src/router/user.ts
RyanL123: this is for testing see testing section

BakerWJ (approved):

PR #161, opened 2023-11-14
[CFD-63] Create Participants Model and Update Absence Model

1zhaohel:
### Description

Created a participant model with one-to-one relation with user and one-to-many relation with the absences model. 

### Reason for Change

Completes CFD-63

### Type of change

### Testing

Tested the participants many-to-many relationship with users as well as logging an absence via participant id.

https://github.com/uoftblueprint/centre-for-dreams/assets/129004013/4fda42ef-c3f3-4be1-a38a-5d66e78fdcf8


packages/db/prisma/schema.prisma
```diff
@@ -13,10 +13,10 @@ datasource db {
 }
 
 model User {
-  id         Int         @id @default(autoincrement())
-  clerkId    String      @unique
-  absences   Absence[]
-  pushTokens PushToken[]
+  id           Int           @id @default(autoincrement())
+  clerkId      String        @unique
+  participants Participant[]
```
BakerWJ, line 18: Each user only needs to be linked to one participant

1zhaohel: wait, i'll edit absence and participant model to use userId

1zhaohel: ok! the ticket said participants should have a many-to-many relationship with user field which is why i did many-to-many relation.

BakerWJ: mb thanks for fixing

packages/api/src/trpc.ts
```diff
@@ -120,6 +120,16 @@ const isAuthed = t.middleware(async ({ next, ctx }) => {
...
 
+  await ctx.db.participant.upsert({
+    where: {
+      id: 1,
+    },
+    update: {},
+    create: {
+      name: "Test",
+    },
+  });
```
BakerWJ, line 131: does this make sense to you @RyanL123

BakerWJ (commented):
lgtm, will let Ryan take a look since he has worked on the auth middleware

packages/api/src/trpc.ts
BakerWJ: basically every user needs to have a linked participant, so for now we just create a dummy participant

BakerWJ: this does make our admin case a bit wonky though.....

BakerWJ: maybe we just have a separate Admin table, don't lump them in with users?

RyanL123: What's the reasoning for separating participants with users if they're a 1-to-1 relationship? Why can we not just re-use the users model?

BakerWJ: They aren't a 1-to-1 relationship -- like there can be 2 parents and one participant

1zhaohel: mb, i think i did a 1-to-1 relationship between User and Participant because that's what i thought we wanted but I can change it so a participant can be linked to many users

BakerWJ: oh I missed that in the schema 🤦 -- pls make that change so that a Participant can have multiple users

1zhaohel: ok 👍

packages/db/prisma/schema.prisma
```diff
@@ -52,11 +53,18 @@ model Activity {
...
+  absenceDate   DateTime    @db.Date
 
-  @@unique([absenceDate, userId])
+  @@unique([absenceDate, participantId])
+}
+
+model Participant {
+  id       Int       @id @default(autoincrement())
+  name     String
+  user     User[]
```
BakerWJ, line 66: 🥲 last change I promise, but should this be users instead?

BakerWJ: for line 66

1zhaohel: good catch, thank you!

packages/db/prisma/schema.prisma
```diff
@@ -52,11 +53,18 @@ model Activity {
...
+  participant   Participant @relation(fields: participantId, references: id)
+  participantId Int
+  absenceDate   DateTime    @db.Date
 
-  @@unique([absenceDate, userId])
+  @@unique([absenceDate, participantId])
+}
+
+model Participant {
+  id       Int       @id @default(autoincrement())
```
RyanL123, line 64: Can we have a comment here on why we can have multiple users mapped to one participant, while each User only maps to a singular participant?

Spoke to Baker and to my understanding the reason is that each User (parent) can have the same kid (Participant), so the state they see for the application should be the same, which is why we need this additional field to link them together.

packages/api/src/trpc.ts
```diff
@@ -120,6 +120,16 @@ const isAuthed = t.middleware(async ({ next, ctx }) => {
     throw new TRPCError({ code: "UNAUTHORIZED" });
   }
 
+  await ctx.db.participant.upsert({
+    where: {
+      id: 1,
+    },
+    update: {},
+    create: {
+      name: "Test",
```
RyanL123, line 129: I don't think this is needed here, since to my understanding a `Participant` presents a child that the `User` is managing (see my comment below).  

In the case of the `upsert` for users we're inserting the user into our database if they aren't there yet. However for participants we don't have to enforce that since maybe the user just signed up and hasn't added a participant in the database yet to link to their account.

BakerWJ: This is just a stopgap fix right now, as every user needs a participant. Thus if a user isn't created in a db, then we also need to have a partipicant assigned to them. When we add the admin functionality of adding whitelisted users, we can remove this logic.

RyanL123: What's our plan for adding participants to a user? E.g. user will go to a dashboard and add their participant? But what if they want to select from existing kids because another user has already created the participant? Will we give them a dropdown list to pick from?

BakerWJ: I think it will be a part of the whitelisting process that the admin will perform -- so they will create a participant and the necessary users

packages/db/prisma/schema.prisma
1zhaohel: ok 👍 just a thought but what if a user (parent) had two participants (kids) in the program :o

BakerWJ: that isn't a case that the NPO has ever had historically -- we can build it out later if necessary, but this will be extra work so will come at the end of the project

1zhaohel: i see, tyty!

BakerWJ (approved):

PR #168, opened 2023-11-15
Use @clerk/nextjs in API Package

BakerWJ:
### Description

Uses `@clerk/nextjs` instead of the Node SDK for the API package.

### Reason for Change

We aren't directly using the Node sdk for our middleware currently.
### Type of change

- [x] Dependency Update


RyanL123 (approved):
LGTM, pulled changes and tested locally, everything works fine

PR #175, opened 2023-11-18
CFD-43: Mobile Login/Signup Page

RyanL123:
### Reason for Change

Completes CFD-43

### Type of change

- [x] Documentation update

### Testing

Tested by loading into Expo Go

Login page
image

Toast showing error 
image

Password reset code
image

After login
image

### DIFF
--- a/apps/expo/src/app/index.tsx
+++ b/apps/expo/src/app/index.tsx
@@ -1,9 +1,15 @@
 import React, { useEffect, useState } from "react";
-import { Text, View } from "react-native";
+import { Button, Text, View } from "react-native";
+import { RootSiblingParent } from "react-native-root-siblings";
 import { SafeAreaView } from "react-native-safe-area-context";
 import * as Notifications from "expo-notifications";
 import { Stack } from "expo-router";
+import { SignedIn, SignedOut, useAuth } from "@clerk/clerk-expo";
 
+import ResetPassword from "~/components/ResetPassword";
+import SignInScreen from "~/components/SignInScreen";
+import SignInWithOAuth from "~/components/SignInWithOAuth";
+import SignUpScreen from "~/components/SignUpScreen";
 import registerForPushNotificationsAsync from "~/notifications/registerNotifications";
 
 Notifications.setNotificationHandler({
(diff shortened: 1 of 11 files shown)


RyanL123:
Not ready for review yet, just want to see whether build succeeds. 

@BakerWJ there's a problem with loading environmental variables in `app.config.ts` for Expo. Since our env variables are in the root directory (the parent directory of Expo), it's not able to load them via `process.env`, but it works if we copy the `.env` file into the `apps/expo`. Interestingly enough Next doesn't care about this, do we have som additional setup? 

Also, `app.config.ts` can read from `process.env` when it's in the context of Github's CI (which is where we're loading `EXPO_PROJECT_ID`), so I'm a little confused on how it's actually loading environmental variables.

apps/expo/src/components/ResetPassword.tsx
```diff
@@ -0,0 +1,125 @@
...
+  clerkError: boolean;
+  errors: {
+    code: string;
+    message: string;
+    longMessage: string;
+    meta: {
+      paramName: string;
+    };
+  }[];
+}
```
BakerWJ, line 17: why is this interface being created manually? can it not be inferred / imported?

apps/expo/src/app/index.tsx
```diff
@@ -27,13 +50,25 @@ const Index = () => {
...
 
   return (
-    <SafeAreaView className="">
-      {/* Changes page title visible on the header */}
-      <Stack.Screen options={{ title: "Home Page" }} />
-      <View className="h-full w-full p-4">
-        <Text className="font-poppins"> Index </Text>
-      </View>
-    </SafeAreaView>
+    <RootSiblingParent>
```
BakerWJ, line 53: should this be in layout instead?

apps/expo/src/app/index.tsx
BakerWJ, line 1: for now, let's remove this from index.tsx -- it seems to be getting a bit bloated with people's testing

apps/expo/src/components/SignUpScreen.tsx
```diff
@@ -0,0 +1,127 @@
+import * as React from "react";
```
BakerWJ, line 1: would recommend not using import *

BakerWJ (commented):
looks good -- for now let's remove the signin stuff from index.tsx, but your next ticket will be building out the signin pages

apps/expo/src/components/ResetPassword.tsx
RyanL123: when you catch an error the type is `unknown`, and to be able to access fields on it you must cast it to some known type. Clerk actually doesn't export this, so we have to manually create it

BakerWJ (approved):
lgtm

PR #178, opened 2023-11-19
[CFD-62] Register push token endpoint

jasonfyw:
### Description

This PR adds an endpoint to the `notification` router that allows a push token to be associated with a user

### Reason for Change

Completes CFD-62

### Type of change

### Testing

None (tRPC panel is broken lol)

### DIFF
--- a/packages/api/src/router/notification.ts
+++ b/packages/api/src/router/notification.ts
@@ -4,7 +4,7 @@ import { z } from "zod";
 
 import { logger } from "@cfd/logger";
 
-import { adminProcedure, createTRPCRouter } from "../trpc";
+import { adminProcedure, createTRPCRouter, protectedProcedure } from "../trpc";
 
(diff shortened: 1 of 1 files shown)


packages/api/src/router/notification.ts
```diff
@@ -45,4 +45,22 @@ export const notificationRouter = createTRPCRouter({
...
+    .input(
+      z.object({
+        pushToken: z.string(),
+      }),
+    )
+    .mutation(async ({ input, ctx }) => {
+      const userId = ctx.userId;
+      if (!userId) {
+        throw new Error("User not found");
+      }
```
BakerWJ, line 58: This check isn't necessary -- the userId should always be available

BakerWJ (commented):
will fix trpc panel for testing

BakerWJ (approved):

PR #184, opened 2023-11-19
Use dotenv for Expo Dev

BakerWJ:
### Description

Uses Dotenv when running any dev commands in the expo folder

### Reason for Change

Environment variables in the root `.env` file were not accesable when using Expo from `pnpm dev`.

### Type of change

### Testing

screenshot

Added "Baker" to `.env`, then read it in with `app.config.ts` and put in the index page


BakerWJ:
@RyanL123 what is the exact command that you use for tunnel, I can add a `pnpm dev:tunnel` command if that is helpful

RyanL123:
> @RyanL123 what is the exact command that you use for tunnel, I can add a `pnpm dev:tunnel` command if that is helpful

`npx expo start --go --tunnel`

RyanL123 (approved):
lgtm, thanks for the fix! Btw is the projectId for expo something we need to keep secret? If not, can we add it to the readme or the example env? Right now when you build the app it's giving an error
``` 
ERROR  [Error: No 'projectId' found. If 'projectId' can't be inferred from the manifest (eg. in bare workflow), you have to pass it in yourself.]
```

ganeshasapu (approved):

PR #185, opened 2023-11-19
CFD-59: Admin dashboard authorization + allow admin to approve users

RyanL123:
### Reason for Change

Completes CFD-59

### Type of change

### Testing

- Clicking the `approve` and `unapprove` buttons correctly updates the records in the database
- If you mark yourself as `isAdmin`, you can access the admin dashboard. Otherwise, you will be redirected back to the home page
- You can only view my information on the main page if you're marked as admin (since it uses adminProcedure), otherwise the information will be blank.

image
What each entry looks on the admin dashboard. Clicking approve/unapprove will alter the approval status in our database accordingly

image
isApproved getting updated in Clerk dashboard


BakerWJ:
you don't need to add the @clerk/backend package, you can use the clerkClient from the nextjs package for this operation: https://clerk.com/docs/references/backend/overview

RyanL123:
> you don't need to add the @clerk/backend package, you can use the clerkClient from the nextjs package for this operation: https://clerk.com/docs/references/backend/overview

For some reason it kept saying `clerkClient.users` is undefined, but that was when I tried it in the frontend and not TRPC, I can try it again.

BakerWJ:
```ts
import { createTRPCRouter, protectedProcedure } from "../trpc";
import { clerkClient } from "@clerk/nextjs";


export const userRouter = createTRPCRouter({
  baker: protectedProcedure.query(async () => {
    return await clerkClient.users.getUser("baker")
  })
});
```
This worked for me and passed type checks

RyanL123:
> ```ts
> import { createTRPCRouter, protectedProcedure } from "../trpc";
> import { clerkClient } from "@clerk/nextjs";
> 
> 
> export const userRouter = createTRPCRouter({
>   baker: protectedProcedure.query(async () => {
>     return await clerkClient.users.getUser("baker")
>   })
> });
> ```
> 
> This worked for me and passed type checks

Interesting, I think it was a problem with using it in the client instead of the server

BakerWJ:
Also some of ur type errors for checks might be beacuse I removed the user router from the root router, so might need to re-add it

apps/nextjs/src/pages/admindashboard.tsx
```diff
@@ -0,0 +1,81 @@
...
+      router.push("/");
+    } else if (currentUserQuery.data?.isAdmin === true) {
+      // user is explicitly an admin, we're no longer waiting for auth to resolve
+      setIsAuthStateLoading(false);
+    }
+  }, [userClerkId, currentUserQuery, isLoaded, user, router]);
+
+  // we're trying to resolve the auth to determine if user has access to this page
+  if (!isLoaded || !user || isAuthStateLoading) return <div>Loading...</div>;
+
```
BakerWJ, line 47: all of this should just be done in the middleware

packages/db/prisma/migrations/20231119232132_add_is_approved_field_to_user/migration.sql
BakerWJ, line 1: delete migrations and just combine into one

BakerWJ (changes requested):
can u put a picture of the new page in the PR description

apps/nextjs/src/pages/admindashboard.tsx
RyanL123: I don't think you can interact with the db in middleware, and putting stuff in `trpc.ts` only protects the trpc routes

BakerWJ: hmmmmmmmmmm, maybe we should use clerk user metadata for admin functionality then?

apps/nextjs/src/pages/admindashboard.tsx
```diff
@@ -0,0 +1,63 @@
...
+  useEffect(() => {
+    // user is not loaded, we're not sure if they're an admin yet
+    if (!isLoaded || !user) {
+      return;
+    }
+
+    if (user.publicMetadata.isAdmin !== true) {
+      router.push("/");
+    }
```
BakerWJ, line 24: can you add this check to the middleware instead?

RyanL123: moved

packages/api/src/router/user.ts
```diff
@@ -1,3 +1,32 @@
...
+  getAllUsers: protectedProcedure.query(async ({ ctx }) => {
+    return await ctx.db.user.findMany();
+  }),
+
+  // Fetches the clerk data of every user that has signed up and returns a mapping of the form:
+  // <clerkId, userData>
+  getAllClerkUsers: protectedProcedure.query(async () => {
+    const users = await clerkClient.users.getUserList();
+    return new Map(users.map((user) => [user.id, user]));
+  }),
```
BakerWJ, line 16: should non-admin users have access to this information?

packages/api/src/router/user.ts
```diff
@@ -1,3 +1,32 @@
...
+import { clerkClient } from "@clerk/nextjs";
+import { z } from "zod";
 
-export const userRouter = createTRPCRouter({});
+import { adminProcedure, createTRPCRouter, protectedProcedure } from "../trpc";
+
+export const userRouter = createTRPCRouter({
+  getAllUsers: protectedProcedure.query(async ({ ctx }) => {
+    return await ctx.db.user.findMany();
+  }),
```
BakerWJ, line 9: same comment -- should non-admins be able to access this?

apps/nextjs/src/middleware.ts
```diff
@@ -1,10 +1,25 @@
...
 // See https://clerk.com/docs/references/nextjs/auth-middleware for more information about configuring your Middleware
 export default authMiddleware({
   // "/" is not protected since that's where we land
   // api routes are not protected since authentication is done via protected routes in trpc.ts
   publicRoutes: ["/", "/(api|trpc)(.*)"],
+  async afterAuth(auth, req) {
+    if (req.nextUrl.pathname === "/admindashboard") {
+      // user not logged in
+      if (!auth.userId) {
+        return NextResponse.redirect(new URL("/signin", req.url));
```
BakerWJ, line 13: use `redirectToSignIn` from Clerk instead

apps/nextjs/src/middleware.ts
```diff
@@ -1,10 +1,25 @@
...
+import { NextResponse } from "next/server";
+import { authMiddleware, clerkClient } from "@clerk/nextjs";
 
 // See https://clerk.com/docs/references/nextjs/auth-middleware for more information about configuring your Middleware
 export default authMiddleware({
   // "/" is not protected since that's where we land
   // api routes are not protected since authentication is done via protected routes in trpc.ts
   publicRoutes: ["/", "/(api|trpc)(.*)"],
+  async afterAuth(auth, req) {
+    if (req.nextUrl.pathname === "/admindashboard") {
```
BakerWJ, line 10: This check can be run for all pages

packages/api/src/router/user.ts
BakerWJ, line 1: maybe make an endpoint for a user to make themselves an admin -- this would be used for development rn. that way we can properly protect our endpoints, and you can make yourself an admin if you need access to the frontend

BakerWJ (changes requested):
after this PR, we should be able to easily implement the admin middleware as well, right?

packages/api/src/router/user.ts
RyanL123: I think it's fine, since under some circumstances we may need all the information even if we're not in the admin dashboard (if we have a drop down that needs show all the users for example)

BakerWJ: I'm not sure if we have any usecases like that right now 🤔  -- would recommend that we keep it admin right now, then just move to protected if we need to in the future

RyanL123:
> after this PR, we should be able to easily implement the admin middleware as well, right?
Already part of this PR since the change isn't huge

packages/api/src/trpc.ts
```diff
@@ -160,4 +195,4 @@ export const protectedProcedure = t.procedure.use(isAuthed);
  * Should be used for all admin dashboard functionality.
  *
  */
-export const adminProcedure = t.procedure.use(isAuthed);
+export const adminProcedure = t.procedure.use(isAdminAuthed);
```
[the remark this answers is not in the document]
RyanL123: didn't know that was possible, changed

packages/api/src/router/user.ts
```diff
@@ -1,3 +1,48 @@
...
+          isApproved: input.isApproved,
+        },
+      });
+    }),
+
+  // This is strictly used for internal testing and should be removed in the future
+  changeAdminStatus: protectedProcedure
+    .input(
+      z.object({
+        clerkId: z.string(),
```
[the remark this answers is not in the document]
RyanL123: I wanted to keep it flexible but I suppose there's no reason to change the admin status of anyone other than the current user. changed

PR #191, opened 2023-11-22
[CFD-71] Activity component

helenaglow:
### Description

Based on figma, created a multi use activity component! 
- Has parameters activity name, attending (whether they accepted or said maybe), and location (name of location).

I wasn't sure if we wanted to include location but it seemed fitting since it looks like the same component.
### Reason for Change

Completes CFD-71

### Type of change

### Testing

```
<Activity activityName="External Events" attending={true/false} location="Canada"} />
```
in `index.tsx`

https://github.com/uoftblueprint/centre-for-dreams/assets/113125436/44ca4ea3-704a-4d43-a6bd-7d04a7b7ef85

### DIFF
--- a/apps/expo/src/app/activity.tsx
+++ b/apps/expo/src/app/activity.tsx
@@ -1,0 +1,57 @@
+import React from "react";
+import { Text, View } from "react-native";
+import { AntDesign } from "@expo/vector-icons";
+
+interface ActivityComponent {
+  activityName: string;
+  location?: string;
+  attending?: boolean;
+}
+
+export default function Activity({
+  activityName,
+  location,
+  attending,
+}: ActivityComponent) {
+  return (
+    <View className="mb-4 w-8/12 flex-col rounded-md bg-zinc-300 p-3">
+      <Text className="font-inter text-base font-semibold text-black">
+        {activityName}
+      </Text>
+
+      {location !== undefined && (
+        <View className="flex-row items-center justify-start gap-2">
+          <AntDesign name="enviromento" size={20} color="black" />
+          <Text className="font-inter text-base font-normal text-black">
+            {location}
+          </Text>
+        </View>
+      )}
+
+      {attending !== undefined && (
+        <View className="flex-row items-center justify-start">
+          <Text className="font-inter text-base font-normal text-black">
+            Attendance:
+          </Text>
+          <View className="ml-3 flex-row items-center justify-start gap-2">
+            {attending ? (
+              <>
+                <AntDesign name="checkcircleo" size={20} color="#22AB00" />
+                <Text className="text-checkGreen font-inter text-base font-normal">
+                  Join
+                </Text>
+              </>
+            ) : (
+              <>
+                <AntDesign name="questioncircleo" size={20} color="black" />
+                <Text className="font-inter text-base font-normal text-black">
+                  Maybe
+                </Text>
+              </>
+            )}
+          </View>
+        </View>
+      )}
+    </View>
+  );
+}
--- a/apps/expo/tailwind.config.ts
+++ b/apps/expo/tailwind.config.ts
@@ -5,4 +5,11 @@ import baseConfig from "@cfd/tailwind-config";
 export default {
   content: ["./src/**/*.{ts,tsx}"],
   presets: [baseConfig],
+  theme: {
+    extend: {
+      colors: {
+        checkGreen: "#22AB00",
+      },
+    },
+  },
 } satisfies Config;


helenaglow:
Actually give me a sec, I am gonna separate some things into different functions to make it easier to read

apps/expo/src/app/activity.tsx
```diff
@@ -0,0 +1,57 @@
+import React from "react";
+import { Text, View } from "react-native";
+import { AntDesign } from "@expo/vector-icons";
+
+interface ActivityComponent {
+  activityName: string;
+  location?: string;
+  attending?: boolean;
+}
```
BakerWJ, line 9: Can you use the Activity type from the `getSchedule` endpoint for the input to the component? You can extract the activity type from the array by doing https://www.benmvp.com/blog/extracting-typescript-types-functions-objects-arrays/. You can get the output for the getSchedule endpoint from https://github.com/uoftblueprint/centre-for-dreams/blob/main/apps/expo/src/utils/api.tsx#L14

helenaglow: Okay got it! Thanks :))

helenaglow: Hi so actually I don't think that will work best in that case (I think idk I could be wrong). The activity component is gonna need the height and is attending parameters which arent in the type returned by schedule, also not all the items are not necessary. 

It could take in a prop with a GetScheduleType Object, is attending, and height. But that will make it a little different if we want the activity component to solely show location or is attending or nothing based on the figma. Unless we want a different Activity component for each version. (Sorry if that doesn't make sense).

Let me know if I misunderstood something tho :)). thank you!!!

helenaglow:
I am still trying to figure out how to do minimum height without hardcoding. I tried `min-h-` but it didnt work as intended :(. Any ideas would be helpful!!!

But other than that issue, I think it can be reviewed.

apps/expo/src/app/activity.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { Text, View } from "react-native";
+import { AntDesign } from "@expo/vector-icons";
+
+// Props specific for the Activity component
+interface ActivityProps {
+  name: string;
+  location?: string;
+  attending?: boolean;
+  activityHeight?: number;
+}
```
BakerWJ, line 11: use RouterOutputs from utils/api.tsx to get the Activity type that is returned by the getSchedule procedure

apps/expo/src/app/activity.tsx
```diff
@@ -0,0 +1,62 @@
+import React from "react";
+import { Text, View } from "react-native";
+import { AntDesign } from "@expo/vector-icons";
+
+// Props specific for the Activity component
+interface ActivityProps {
+  name: string;
+  location?: string;
+  attending?: boolean;
+  activityHeight?: number;
```
BakerWJ, line 10: should height be calculated from the duration of the activity?

apps/expo/tailwind.config.ts
```diff
@@ -5,4 +5,11 @@ import baseConfig from "@cfd/tailwind-config";
 export default {
   content: ["./src/**/*.{ts,tsx}"],
   presets: [baseConfig],
+  theme: {
+    extend: {
+      colors: {
+        checkGreen: "#22AB00",
+      },
+    },
+  },
```
BakerWJ, line 14: wouldn't recommend putting this in theme, as we'll be adding the actual theme values from our design system soon

apps/expo/src/app/activity.tsx
```diff
@@ -0,0 +1,62 @@
+import React from "react";
+import { Text, View } from "react-native";
+import { AntDesign } from "@expo/vector-icons";
```
BakerWJ, line 3: for now you can just use placeholders like "?" for icons, since we will be updating these components with the design system

BakerWJ (changes requested):

helenaglow:
Thank you for reviewing! I will get back to these comments in a bit, I have to work on assignment rn. I just wanted to let you know that I saw them!

apps/expo/tailwind.config.ts
helenaglow: That makes sense!

apps/expo/src/app/activity.tsx
helenaglow: Yea. Thats what I was doing at first but thought that the height might be different based on situation (cause I saw on figma there were some that were varying sizes on different pages) so I was just gonna let the other screens decide. 

Another thing is that size is dependent on duration and line height. Since line height is decided in the schedule page, I didnt want to set it in both places and make it annoying to change in the future. 

Should I figure out a different way?

helenaglow:
https://github.com/uoftblueprint/centre-for-dreams/assets/113125436/97d65b0f-914e-410e-869e-680be5c8ffc4

Updated testing ^^

Changes:
- made the activity component use the router output
- removed icons
- removed the theme color that was previously added

apps/expo/src/app/activity.tsx
helenaglow: If you think it's best to just add line height here, calculate it here, and have a book that dictates it came from schedule, then I can do that. Just let me know your thoughts!

BakerWJ: It might be better to just have this component fill the parent div, and control its size with the parent div instead?

helenaglow: Ohhh interesting! I will do that! Thanks 🙏

helenaglow: Fixed

BakerWJ (approved):

PR #193, opened 2023-11-23
[CFD-29] Add comment procedures

jasonfyw:
### Description

This PR adds:
- Comment router with CRUD for comments under posts
- A user field to announcement and discussion models for the author of the post

### Reason for Change

Completes CFD-29

### Type of change

### Testing

Comment endpoints and modified post models work after basic testing through tRPC panel

### DIFF
--- a/packages/api/src/router/comment.ts
+++ b/packages/api/src/router/comment.ts
@@ -12,27 +12,24 @@ export const commentRouter = createTRPCRouter({
       }),
     )
     .query(async ({ input, ctx }) => {
-      try {
-        return await ctx.db.comment.findMany({
-          where: { postId: input.postId },
-          orderBy: { createdAt: "desc" },
-          select: {
-            id: true,
-            text: true,
-            createdAt: true,
-          },
+      const post = await ctx.db.post.findUnique({
+        where: { id: input.postId },
+      });
+      if (!post) {
+        throw new TRPCError({
+          code: "NOT_FOUND",
+          message: "Post not found",
         });
-      } catch (e) {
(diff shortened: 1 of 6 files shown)


packages/api/src/router/comment.ts
```diff
@@ -0,0 +1,123 @@
...
+        id: z.number().nonnegative(),
+        text: z.string(),
+      }),
+    )
+    .mutation(async ({ input, ctx }) => {
+      const userId = ctx.userId;
+      const comment = await ctx.db.comment.findUnique({
+        where: { id: input.id },
+        select: { id: true, userId: true },
+      });
```
BakerWJ, line 78: will this throw an error if the comment doesn't exist?

BakerWJ (commented):
looks good overall -- just have the one question

packages/api/src/router/comment.ts
jasonfyw: yeah right after i check `if (!comment)`

BakerWJ: i mean like it might throw a prisma error?

jasonfyw: no, it just returns null:
image
a prisma error is only thrown when connecting to an entity that doesn't exist

jasonfyw: also i changed error handling for `comment.list` so that it correctly throws 404 when the provided `postId` does not exist (before it would just return an empty list)

BakerWJ (approved):

PR #199, opened 2023-11-25
[CFD-64] Update Absence Endpoints

1zhaohel:
### Description

Updated `absence.ts` to find user by `ctx.userId`. Throws error if user isn't found.

### Reason for Change

Completes CFD-64

### Type of change

### DIFF
--- a/packages/api/src/router/absence.ts
+++ b/packages/api/src/router/absence.ts
@@ -15,10 +15,17 @@ export const absenceRouter = createTRPCRouter({
       if (!user) {
         throw new TRPCError({ code: "NOT_FOUND" });
       }
-      await ctx.db.absence.create({
-        data: {
-          participantId: user.participantId,
+      await ctx.db.absence.upsert({
+        create: {
           absenceDate: input.absenceDate,
+          participantId: user.participantId,
+        },
+        update: {},
+        where: {
+          absenceDate_participantId: {
+            absenceDate: input.absenceDate,
+            participantId: user.participantId,
+          },
         },
       });
     }),
--- a/packages/api/src/router/absence.ts
+++ b/packages/api/src/router/absence.ts
@@ -11,7 +11,7 @@ export const absenceRouter = createTRPCRouter({
       }),
     )
     .mutation(async ({ input, ctx }) => {
-      const user = await ctx.db.user.findFirst({ where: { id: ctx.userId } });
+      const user = await ctx.db.user.findUnique({ where: { id: ctx.userId } });
       if (!user) {
         throw new TRPCError({ code: "NOT_FOUND" });
       }
@@ -28,7 +28,7 @@ export const absenceRouter = createTRPCRouter({
   }),
 
   getAbsences: protectedProcedure.query(async ({ ctx }) => {
-    const user = await ctx.db.user.findFirst({ where: { id: ctx.userId } });
+    const user = await ctx.db.user.findUnique({ where: { id: ctx.userId } });
     if (!user) {
       throw new TRPCError({ code: "NOT_FOUND" });
     }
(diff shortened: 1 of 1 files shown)


packages/api/src/router/absence.ts
```diff
@@ -6,14 +7,17 @@ export const absenceRouter = createTRPCRouter({
   createAbsence: protectedProcedure
     .input(
       z.object({
-        participantId: z.number().nonnegative(),
         absenceDate: z.date(),
       }),
     )
     .mutation(async ({ input, ctx }) => {
+      const user = await ctx.db.user.findFirst({ where: { id: ctx.userId } });
```
BakerWJ, line 14: i think findUnique makes a bit more sense here, even if they effectively do the same thing

packages/api/src/router/absence.ts
```diff
@@ -23,31 +27,32 @@ export const absenceRouter = createTRPCRouter({
...
 
   deleteAbsence: protectedProcedure
     .input(
       z.object({
-        participantId: z.number().nonnegative(),
         absenceDate: z.date(),
       }),
     )
     .mutation(async ({ input, ctx }) => {
+      const user = await ctx.db.user.findFirst({ where: { id: ctx.userId } });
```
BakerWJ, line 49: findUnique

packages/api/src/router/absence.ts
```diff
@@ -23,31 +27,32 @@ export const absenceRouter = createTRPCRouter({
...
-    )
-    .query(async ({ input, ctx }) => {
-      return await ctx.db.absence.findMany({
-        where: {
-          participantId: input.participantId,
-        },
-      });
-    }),
+  getAbsences: protectedProcedure.query(async ({ ctx }) => {
+    const user = await ctx.db.user.findFirst({ where: { id: ctx.userId } });
```
BakerWJ, line 31: findUnique

packages/api/src/router/absence.ts
```diff
@@ -6,14 +7,17 @@ export const absenceRouter = createTRPCRouter({
...
+      if (!user) {
+        throw new TRPCError({ code: "NOT_FOUND" });
+      }
       await ctx.db.absence.create({
         data: {
-          participantId: input.participantId,
+          participantId: user.participantId,
           absenceDate: input.absenceDate,
         },
       });
```
BakerWJ, line 23: should we make this an upsert instead (not related to ticket but might be necessary)

packages/api/src/router/absence.ts
```diff
@@ -23,31 +27,32 @@ export const absenceRouter = createTRPCRouter({
...
     .mutation(async ({ input, ctx }) => {
+      const user = await ctx.db.user.findFirst({ where: { id: ctx.userId } });
+      if (!user) {
+        throw new TRPCError({ code: "NOT_FOUND" });
+      }
       await ctx.db.absence.deleteMany({
         where: {
-          participantId: input.participantId,
+          participantId: user.participantId,
           absenceDate: input.absenceDate,
```
BakerWJ, line 56: I don't think that this needs to be a deleteMany, as there should only be one row in the database with the given participantId and absenceDate (also not directly related to ticket but should probably be fixed)

BakerWJ (commented):
looks good -- just some really small changes before approving

packages/api/src/router/absence.ts
1zhaohel: do we use upsert so that if they find a record that already has the `absenceDate` and `participantId`, then it would do nothing?

BakerWJ: yep -- I think that if we don't use an upsert, then there will be an error if we try to create one that already exists

1zhaohel: :o ok thank you!

PR #200, opened 2023-11-25
Create workflow for pnpm cache

BakerWJ:
### Description

Creates a workflow that runs on the `main` branch and caches all dependencies.

### Reason for Change

Github Actions will only restore caches from the `main` and current branch. Now that we have a merge queue, we aren't caching our dependencies on pushes to main

### Type of change

- [x] CI

### Testing

Hopefully it works


ganeshasapu (approved):
infra lord

PR #201, opened 2023-11-25
Implemented component for announcement summaries

Ram-Raghav-S:
### Description

Added a component for announcement summaries in the Expo app.

### Reason for Change

CFD-68

Contributes to CFD-

### Type of change

### Testing

- [X] Test A:
-  WhatsApp Image 2023-11-25 at 4 09 01 PM

Code for generating test in `index.tsx`: 
`    <SafeAreaView className="bg-white p-4">
      {/* Changes page title visible on the header */}
      <Stack.Screen options={{ title: "Home Page" }} />
      <View className="w-ful h-full p-2">
        <AnouncementTile
          title="2023 Markham Santa Clause Parade"
          createdBy="Michelle C."
          createdAt={new Date("2023-10-28")}
          titleImageSrc={require("./title_image_placeholder.jpg")}
          userImageSrc={require("./male_user.png")}
        />
      </View>
    </SafeAreaView>

### How it looks like after the changes:

WhatsApp Image 2023-12-03 at 1 08 58 PM

`


apps/expo/src/components/AnnouncementTile.tsx
```diff
@@ -0,0 +1,55 @@
+import { Image, Text, View } from "react-native";
+import type { ImageSourcePropType } from "react-native";
+
+interface Props {
+  title: string;
+  titleImageSrc: ImageSourcePropType;
+  createdAt: Date;
+  createdBy: string;
+  userImageSrc: ImageSourcePropType;
+}
```
BakerWJ, line 10: Use the RouterOutputs helper in utils/api.tsx to infer the type of an Announcement from the getAnnouncements endpoint

apps/expo/src/components/AnnouncementTile.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+function AnouncementTile({
+  title,
+  titleImageSrc,
+  createdAt,
+  createdBy,
+  userImageSrc,
```
BakerWJ, line 17: For now, you can use placeholders for the image sources. They will be added to announcements in a future sprint.

BakerWJ (changes requested):
looks good -- just some small changes

apps/expo/src/components/AnnouncementTile.tsx
Ram-Raghav-S: I'm not sure how to use it because the getAnnouncement endpoint has different types than what I am using. For example, I need createdBy to be a user name, but the endpoint only has the userID for the corresponding user, which is a number. Is there a way to go about this in TypeScript?

apps/expo/src/components/AnnouncementTile.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import type { RouterOutputs } from "~/utils/api";
+
+type GetAnnouncementOutput =
+  RouterOutputs["announcement"]["getAnnouncements"][number];
+
+const AnnouncementTile: React.FC<GetAnnouncementOutput> = ({
+  title,
+  createdAt,
+}) => {
```
BakerWJ, line 13: I think you need to wrap the GetAnnouncementOutput in an interface so that you can just use the component like `<AnnouncementTile announcement={a} />`. Example: 
```javascript
interface AnnouncementTileProps {
    announcement: GetAnnouncementOutput
}

BakerWJ (approved):

PR #173, opened 2023-11-18
Sarinali/cfd 67 comment component

sarinali:
### Description
Created comment component on expo app

Completes CFD-67

- [x] New feature (non-breaking change which adds functionality)

IMG_5746

Test:

/src/apps/expo/app/index.tsx:

```import type { ImageSourcePropType } from "react-native";";

const imageUrl = 'https://static.wikia.nocookie.net/acc-official-database/images/9/91/El_gato.jpg/revision/latest?cb=20220709001857';
const imageSourceProp: ImageSourcePropType = { uri: imageUrl };
<Comment user="Sarina" user_pic={imageSourceProp} time_elapsed="10h" comment_body="thank you so much for merging my PR so quickly i love my PLs and they are awesome #ilovebakerandganesh"></Comment>```

PR #211, opened 2023-11-28
Cfd 67 Comment Component

sarinali:
Created the comment component and fixed the git issues related

### Description
Fixed the git issue 28 files changes (only 2 now) and removed the icon dependency

### Reason for Change
cfd 67

https://github.com/uoftblueprint/centre-for-dreams/pull/173 same code as other PR

IMG_5781


pnpm-lock.yaml
BakerWJ, line 1: did u edit this manually? not sure why it got removed from the expo dependencies

BakerWJ:
also can u update the pr title and description to fit the pr?

pnpm-lock.yaml
BakerWJ: Yeah just revert those changes -- you never want to manually edit a lockfile. In this case the vector icons package is a transient dependency, so we can't just remove entirely (pnpm handles this for us, so we only manage `package.json`)

BakerWJ (approved):

PR #212, opened 2023-11-28
[CFD-74] Add Poppins font to expo

jasonfyw:
### Description
Adds and imports Poppins Google font to Expo. Also creates CSS classes for all typography presets outlined in the Figma using Tailwind plugins

### Reason for Change

Completes CFD-74 and CFD-76

### Type of change

### Testing

Text on top contains Poppins font, text on bottom has default font


apps/expo/src/app/index.tsx
BakerWJ, line 1: I think that we might want to be loading the fonts in _layout.tsx instead: https://stackoverflow.com/questions/74821420/how-to-use-a-custom-font-in-react-native-with-nativewind

jasonfyw: i moved loading the fonts to _layout and created a theme.ts file so the fonts can be used in a similar way to the linked SO answer

apps/expo/src/app/_layout.tsx
```diff
@@ -1,12 +1,18 @@
...
 
 import { TRPCProvider } from "~/utils/api";
 
 // This is the main layout of the app
 // It wraps your pages with the providers they need
 const RootLayout = () => {
+  const [fontsLoaded, fontError] = useFonts({
+    Poppins_400Regular,
+  });
+  if (!fontsLoaded && !fontError) return null;
```
BakerWJ, line 14: this should be replaced by the splashscreen in the future @nm-le

apps/expo/tailwind.config.ts
```diff
@@ -5,4 +5,11 @@ import baseConfig from "@cfd/tailwind-config";
 export default {
   content: ["./src/**/*.{ts,tsx}"],
   presets: [baseConfig],
+  theme: {
+    extend: {
+      fontFamily: {
+        poppins: ["Poppins_400Regular", "sans-serif"],
```
BakerWJ, line 11: since we're only using poppins 400 and poppins 500, maybe the best way to get around the font weight issue is just to have a font-poppins500 and a font-poppins400?

BakerWJ (approved):

PR #219, opened 2023-11-29
[CFD-75] Add color palette to tailwind theme

jasonfyw:
### Description

Adds the colour design system specified in the Figma to the Tailwind theme

### Reason for Change

Completes CFD-75

### Type of change

- [ ] New feature (non-breaking change which adds functionality)

### Testing

Colours are accessible through tailwind classes (e.g. `bg-tertiary-50`)


BakerWJ:
Can you shorten the names to just the first letter, so primary-100 = p-100. I think that's also what is done for some of the color comparisons in the figma

BakerWJ:
also if neutral variant is the same, you can just delete it -- I'm not entirely sure what the purpose of it is

jasonfyw:
@BakerWJ neutral variant had tiny differences for some of the colours – basically the same so i deleted it

BakerWJ (approved):
going to assume these colors are correct -- I won't be the one to check 😅

PR #306, opened 2024-01-10
[CFD-78] Filled Button Components

1zhaohel:
### Description

Created Button/Filled and Button/Disabled based on the Figma design which also accepts an `onClick` prop that passes a function to the underlying Button.

### Reason for Change

Completes CFD-78

### Type of change

### Testing

- Tested the button in `index.tsx`
image


1zhaohel:
not sure if the styling is correct...

apps/nextjs/src/components/filledbutton.tsx
```diff
@@ -0,0 +1,41 @@
...
+  onClick,
+  disabled = false,
+}) => {
+  if (disabled) {
+    return (
+      <button>
+        <div className="inline-flex h-10 w-auto flex-col items-center justify-center gap-2 rounded-full bg-zinc-900 bg-opacity-10">
+          <div className="inline-flex shrink grow basis-0 items-center justify-center gap-2 self-stretch px-[24px] py-[10px]">
+            <div className="font-poppins text-center text-base font-medium tracking-normal text-zinc-900 opacity-40">
+              {label}
```
BakerWJ, line 20: instead of using a label, it might be better to use the children props so it would be like `<Button>Click</Button>`

apps/nextjs/src/components/filledbutton.tsx
BakerWJ, line 1: This button should be built for the React Native app, not the Nextjs app. You should be able to reuse a lot of the css when making it for RN

BakerWJ (changes requested):

BakerWJ (commented):
the PR itself looks good, but it looks like you messed up your git which is why a bunch of files are being changed. I will approve after this is fixed

1zhaohel:
> the PR itself looks good, but it looks like you messed up your git which is why a bunch of files are being changed. I will approve after this is fixed

ok 👍

PR #345, opened 2024-01-21
Ram raghav s/cfd 87 update announcement bubble component

Ram-Raghav-S:
### Description

Updated the announcement card design to use hi-fi designs.

Note that I have left a few placeholders in the code, which will be removed once their is a way to retrieve the corresponding info.

### Reason for Change

CFD-87

Contributes to CFD-

### Type of change

### Testing

- [X] Test A: 
WhatsApp Image 2024-01-21 at 11 28 20 AM

Code for generating test that goes in index.tsx: 
`      <View className="h-full bg-white px-[20px]">
        <AnnouncementCard
          announcement={{
            id: 0,
            title: "2023 Markham Santa Claus Parade",
            contents: "",
            createdAt: new Date(),
            postType: "Announcement",
            userId: 0,
          }}
        />
      </View>
`

### DIFF
--- a/apps/expo/src/components/AnnouncementCard.tsx
+++ b/apps/expo/src/components/AnnouncementCard.tsx
@@ -38,10 +38,7 @@ const AnnouncementCard: React.FC<AnnouncementCardProps> = ({
         />
       </View>
       <View className="flex-[0.7] gap-4 p-4 pl-0">
-        <Text
-          numberOfLines={1}
-          className="text-base font-medium tracking-[0.15px]"
-        >
+        <Text numberOfLines={1} className="font-title-md">
           {title}
         </Text>
         <View className="flex-row justify-between">
@@ -52,9 +49,9 @@ const AnnouncementCard: React.FC<AnnouncementCardProps> = ({
               resizeMode="cover"
               className="outline-3 h-5 w-5 outline-black"
             />
-            <Text className="text-sm">{createdBy}</Text>
+            <Text className="font-body-m">{createdBy}</Text>
           </View>
-          <Text className="text-sm">{createdAtFormatted}</Text>
+          <Text className="font-body-m">{createdAtFormatted}</Text>
         </View>
       </View>
       <View className="w-1 flex-[0.05]"></View>
--- a/apps/expo/src/components/AnnouncementCard.tsx
+++ b/apps/expo/src/components/AnnouncementCard.tsx
@@ -6,11 +6,11 @@ import type { RouterOutputs } from "~/utils/api";
 type GetAnnouncementOutput =
   RouterOutputs["announcement"]["getAnnouncements"][number];
 
-interface AnnouncementTileProps {
+interface AnnouncementCardProps {
   announcement: GetAnnouncementOutput;
 }
 
-const AnnouncementTile: React.FC<AnnouncementTileProps> = ({
+const AnnouncementCard: React.FC<AnnouncementCardProps> = ({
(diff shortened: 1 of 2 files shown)


apps/expo/src/components/AnnouncementCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+          resizeMode="cover"
+        />
+      </View>
+      <View className="flex-[0.7] gap-4 p-4 pl-0">
+        <Text
+          numberOfLines={1}
+          className="text-base font-medium tracking-[0.15px]"
+        >
+          {title}
+        </Text>
```
BakerWJ, line 46: right now, the title gets cut off after the first line. Can we make it so that it can expand to maybe 2 or 3 lines if the title is long enough?

apps/expo/src/components/AnnouncementCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+              accessibilityLabel={`${createdBy}'s picture`}
+              resizeMode="cover"
+              className="outline-3 h-5 w-5 outline-black"
+            />
+            <Text className="text-sm">{createdBy}</Text>
+          </View>
+          <Text className="text-sm">{createdAtFormatted}</Text>
+        </View>
+      </View>
+      <View className="w-1 flex-[0.05]"></View>
```
BakerWJ, line 60: is this view doing anything?

apps/expo/src/components/AnnouncementCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+          <View className="flex-row items-center gap-[5px]">
+            <Image
+              source={userImageSrc}
+              accessibilityLabel={`${createdBy}'s picture`}
+              resizeMode="cover"
+              className="outline-3 h-5 w-5 outline-black"
+            />
+            <Text className="text-sm">{createdBy}</Text>
+          </View>
+          <Text className="text-sm">{createdAtFormatted}</Text>
```
BakerWJ, line 57: I didn't know that you could use text-sm or text-base and it would still work -- would recommend using the utility classes from #212 for consistency

BakerWJ (commented):
looks good -- just some small nits

apps/expo/src/components/AnnouncementCard.tsx
Ram-Raghav-S: Sure, I can do that. But it wouldn't look like the figma design, is that ok?
image

apps/expo/src/components/AnnouncementCard.tsx
Ram-Raghav-S: Just acts as flexible padding in the right. I can try doing it another way too if you'd like

apps/expo/src/components/AnnouncementCard.tsx
Ram-Raghav-S: Ah alright. I will make the change

apps/expo/src/components/AnnouncementCard.tsx
BakerWJ: you can keep it how it is for now then -- we can bring this up with the designers later

apps/expo/src/components/AnnouncementCard.tsx
Ram-Raghav-S: Done. I have added updated the testing screenshot with the new image.

BakerWJ (approved):

PR #347, opened 2024-01-21
[CFD-81] Create tab component

jasonfyw:
### Description
Adds a tab nav component to Expo according to the Figma design

### Reason for Change

Completes CFD-81

### Type of change

### Testing

Example usage:

```typescript
<TabNav>
    <TabNav.Tab label="Tab 1" onPress={() => console.log("switched to tab 1")} />
    <TabNav.Tab label="Tab 2" onPress={() => console.log("switched to tab 2")} />
    <TabNav.Tab label="Tab 3" onPress={() => console.log("switched to tab 3")} />
</TabNav>
```


apps/expo/src/components/TabNav/TabNav.tsx
```diff
@@ -0,0 +1,31 @@
+import type { ReactElement } from "react";
+import React, { useState } from "react";
+import { View } from "react-native";
+import { styled } from "nativewind";
+
+import Tab from "./Tab";
+
+const StyledView = styled(View);
```
BakerWJ, line 8: views should be styled by default, so I don't think this is necessary?

BakerWJ (approved):
Looks good!

apps/expo/src/components/TabNav/Tab.tsx
BakerWJ, line 1: I think it may be better to accept the label through a child component than through a label prop -- in one of the designs right now we're using an SVG as a label, which I don't think is compatible with this design:

image

jasonfyw: Ok done. Example:

image

BakerWJ (approved):

PR #348, opened 2024-01-21
[CFD-78] Filled Button Components

1zhaohel:
### Description

Created Filled and Disabled Button based on the Figma design which also accepts an `onClick` prop that passes a function to the underlying Button.

### Reason for Change

Completes CFD-78

### Type of change

### Testing

- Tested the button in `index.tsx`
```
import FilledButton from "~/components/FilledButtons";
function sayHello() {
  alert("You clicked me!");
}
```
place in return
```
<FilledButton onClick={sayHello} disabled={true}>
  Label
</FilledButton>
<FilledButton onClick={sayHello} disabled={false}>
  Label
</FilledButton>
```
image

### DIFF
--- a/apps/expo/src/components/FilledButtons.tsx
+++ b/apps/expo/src/components/FilledButtons.tsx
@@ -12,61 +12,31 @@ const FilledButton: React.FC<FilledButtonProps> = ({
   disabled = false,
   children,
 }) => {
-  return (
-    <TouchableOpacity onPress={onClick} disabled={disabled}>
-      <View style={disabled ? styles.disabledButton : styles.enabledButton}>
-        <Text
-          style={disabled ? styles.disabledText : styles.enabledText}
-          className="font-poppins"
-        >
-          {children}
-        </Text>
-      </View>
-    </TouchableOpacity>
-  );
+  if (disabled) {
+    return (
+      <TouchableOpacity onPress={onClick} disabled={disabled}>
+        <View className="inline-flex h-12 items-center justify-center gap-2">
+          <View className="bg-n-10/10 rounded-full px-6 py-2.5">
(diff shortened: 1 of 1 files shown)


BakerWJ (changes requested):
buttons look good -- can you convert the styling to use tailwind? we have utility classes for colors and font from #212 and #219

apps/expo/src/components/FilledButtons.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const FilledButton: React.FC<FilledButtonProps> = ({
+  onClick,
+  disabled = false,
+  children,
+}) => {
+  if (disabled) {
+    return (
+      <TouchableOpacity onPress={onClick} disabled={disabled}>
+        <View className="inline-flex h-12 items-center justify-center gap-2">
```
BakerWJ, line 18: The button shouldn't have any css that affects how its parent might use it -- for the screenshot you sent, it should be on the left of the screen directly under the text if there is no styling in the parent view

apps/expo/src/components/FilledButtons.tsx
```diff
@@ -0,0 +1,42 @@
...
+              {children}
+            </Text>
+          </View>
+        </View>
+      </TouchableOpacity>
+    );
+  }
+};
+
+export default FilledButton;
```
BakerWJ, line 42: I think we should wrap the export with Nativewind's styled (https://www.nativewind.dev/api/styled) so that users can modify the parent container of the button if necessary

apps/expo/src/components/FilledButtons.tsx
1zhaohel: okay sounds good mb!

BakerWJ (approved):

PR #351, opened 2024-01-27
[CFD-86] Outlined Button Component

1zhaohel:
### Description

Created two of the three outlined buttons. The third outlined button will have to take in an icon so waiting for a PR. 

### Reason for Change

Completes CFD-86

### Type of change

### Testing

Tested the two buttons in `index.tsx`
image
image
image

```
<View className="flex flex-1 flex-col items-center justify-start">
  <View className="w-30">
    <OutlinedButton onPress={sayHello} fill={false}>
      Label
    </OutlinedButton>
  </View>
</View>
```


apps/expo/src/components/OutlinedButtons.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import { Text, TouchableOpacity, View } from "react-native";
+
+interface OutlinedButtonProps {
+  onClick: () => void;
```
BakerWJ, line 5: change this to onPress for consistency

BakerWJ (commented):
the text in the image doesn't look centered in the screenshot, so may want to confim that it is centered

1zhaohel:
> the text in the image doesn't look centered in the screenshot, so may want to confim that it is centered

i see it too - to change i made the vertical padding 8px instead of 10px (from figma) since 8 looks more centered?

apps/expo/src/components/OutlinedButtons.tsx
```diff
@@ -0,0 +1,31 @@
...
+const OutlinedButton: React.FC<OutlinedButtonProps> = ({
+  onPress,
+  children,
+  fill,
+}) => {
+  return (
+    <TouchableOpacity onPress={onPress}>
+      <View className="h-12 flex-auto items-start gap-2">
+        <View
+          className={`border-p-40 rounded-full border bg-white px-6 py-2 ${
```
BakerWJ, line 20: is there a better way to center the text than by using padding?

1zhaohel: I tried a lot of methods like `items-center`, `content-center`, messing with gap and width, structuring the views differently, but couldn't find any other way without the button losing its structure TT. Do you know any way to solve this issue?

BakerWJ: ok thats fine, as long as you tried other options

BakerWJ (approved):

apps/expo/src/components/OutlinedButtons.tsx
BakerWJ, line 20: remove nested views

BakerWJ (changes requested):
take a look at the changes in #355 -- these nested views seem to be redundant for the button

apps/expo/src/components/OutlinedButtons.tsx
1zhaohel: okay sounds good! should i also change how FilledButtons is structured?

BakerWJ: Ryan already made changes to those ones, so you can just modify the Outlined one

apps/expo/src/components/OutlinedButtons.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import { Text, TouchableOpacity } from "react-native";
+
+interface OutlinedButtonProps {
+  onPress: () => void;
+  children: React.ReactNode;
+  fill: boolean;
+  // icon to be added
```
BakerWJ, line 8: why does the icon need to be added? it should be able to be inserted as a child like `<Button>Baker <Icon /> </Button`

BakerWJ (approved):

PR #352, opened 2024-02-01
[CFD-79] Create input box components

jasonfyw:
### Description
Adds a flexible and customizable text input component based on React Native's `TextInput`

### Reason for Change

Completes CFD-79

### Type of change

- [x] New feature (non-breaking change which adds functionality)

### Testing

Tested single-line input, multiline input, different input modes (email, numeric, password, etc.).

Example code:

```typescript
import Ionicons from "@expo/vector-icons/Ionicons";
import { TextInput } from "~/components/inputs";

<TextInput
    placeholder="Write description here..."
    multiline
    className="my-2 h-[168px]"
/>
<View>
  <TextInput placeholder="Search" inputMode="numeric" multiline>
      <TextInput.RightElement>
          <Ionicons
              name="search"
              size={24}
              color="black"
              onPress={() => alert("hi there")}
          />
      </TextInput.RightElement>
  </TextInput>
</View>
<TextInput
    placeholder="Enter email"
    multiline
    className="my-2"
    inputMode="email"
/>
```

PR #355, opened 2024-02-03
CFD-89: Mobile login page

RyanL123:
### Reason for Change

Fixes CFD-78

Completes CFD-89

### Type of change

### Testing

image


apps/expo/src/components/NavBar.tsx
```diff
@@ -73,6 +73,13 @@ const NavBar = () => {
           tabBarLabel: "Account",
         }}
       />
+      <Tabs.Screen
+        name="signin"
+        options={{
+          href: "/signin",
+          tabBarLabel: "Sign In",
+        }}
+      />
```
BakerWJ, line 82: Instead of creating a Tab, I think that you want to wrap the NavBar in _layout.tsx with the SignedIn component: https://clerk.com/docs/quickstarts/expo#protecting-your-pages

apps/expo/src/app/signin.tsx
```diff
@@ -0,0 +1,37 @@
...
+      </View>
+      <View className="flex-1 justify-center">
+        <View className="gap-2">
+          <Text aria-label="Label for Email">Email</Text>
+          <TextInput
+            placeholder="Enter Email"
+            className="w-100 rounded-lg border p-2"
+            aria-label="input"
+          />
+        </View>
```
BakerWJ, line 26: would recommend refactoring the input box to use #352 after it is merged in

BakerWJ (changes requested):

apps/expo/src/app/signin.tsx
RyanL123: I'll wait for that one to go in first

apps/expo/src/app/signin.tsx
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import { Image, Text, TextInput, View } from "react-native";
+import type { ImageSourcePropType } from "react-native";
+
+import FilledButton from "~/components/FilledButtons";
+import Logo from "../../assets/logo.png";
```
RyanL123, line 6: Should we do this with absolute imports? Not sure what the best method is here but `~` is relative to the `src` folder, so we'd have to do `~/../assets` which feels a little weird

BakerWJ: I think that what you have is fine, as long as it works I don't think it's too important

apps/expo/src/components/FilledButtons.tsx
```diff
@@ -12,27 +12,19 @@ const FilledButton: React.FC<FilledButtonProps> = ({
...
-      </TouchableOpacity>
-    );
-  }
+  const bgColor = disabled ? "bg-n-10/10" : "bg-p-40";
+  const textColor = disabled ? "text-n-10/40" : "text-e-100";
+  return (
+    <TouchableOpacity
+      onPress={onClick}
+      disabled={disabled}
+      className={`${bgColor} h-12 justify-center rounded-full px-6 py-2.5`}
```
1zhaohel, line 21: btw without the gap css, i think u need to make `h-10` (40 px)

RyanL123: `h-10` would make the text kind of "sink into" the button, I think it's fine without the gap. If we notice a problem we can address in a follow up PR

apps/expo/src/components/SignInScreen.tsx
RyanL123, line 1: This is replaced with `signin.tsx`

apps/expo/src/components/NavBar.tsx
```diff
@@ -73,6 +73,12 @@ const NavBar = () => {
           tabBarLabel: "Account",
         }}
       />
+      <Tabs.Screen
+        name="signin" // the signin screen should not be shown as a tab
+        options={{
+          href: null,
+        }}
+      />
```
BakerWJ, line 81: remove

apps/expo/src/app/_layout.tsx
```diff
@@ -31,8 +32,13 @@ const RootLayout = () => {
...
       */}
-          <NavBar />
-          <StatusBar />
+          <SignedOut>
+            <SignInScreen />
+          </SignedOut>
+          <SignedIn>
+            <NavBar />
+            <StatusBar />
+          </SignedIn>
```
BakerWJ, line 41: can you revert this logic until we get working passwordless, so that devs don't need to worry about creating an account

apps/expo/src/components/NavBar.tsx
RyanL123: Why? This is needed to hide the tab at the bottom

BakerWJ (approved):
looks great -- will reapprove once conflict has been resolved :)

BakerWJ (approved):

PR #357, opened 2024-02-07
Fix `renovate.json`

BakerWJ:
### Description

Fixes scheduling for dependency updates

### Reason for Change

Renovate wasn't working because of invalid settings

### Type of change

- [x] Other


ganeshasapu (approved):
rennovate is back

PR #360, opened 2024-02-11
Set Cocoapods version in Github Actions

BakerWJ:
### Description

Adds a setup step for Cocoapods for Github Actions

### Reason for Change

https://github.com/CocoaPods/CocoaPods/issues/12226


ganeshasapu (approved):
ci/cd king

PR #367, opened 2024-02-17
[CFD-85] Outlined Chip Components

1zhaohel:
### Description

Created outlined chip components. Added two colours to `tailwind.config.ts`, lmk if I should change how it's formatted. 

### Reason for Change

Completes CFD-85

### Type of change

### Testing

In `index.tsx`:
 ```
<View className="flex flex-1 flex-col items-start justify-start px-3 py-2">
  <View className="">
    <OutlinedChip onPress={sayHello} disabled={true}>
      Maybe
    </OutlinedChip>
  </View>
</View>
```
image
image
image
image
image
image


apps/expo/src/components/OutlinedChips.tsx
BakerWJ, line 1: I think it is better if you create separate components for each chip -- the logic with checking the content of the children isn't necessary

BakerWJ (changes requested):
they look good -- hopefully should just be some copy and paste to create each of them

apps/expo/src/components/OutlinedChips.tsx
1zhaohel: ok 👍

PR #368, opened 2024-02-17
[CFD-102] Re-implement schedule data model

jasonfyw:
### Description

Creates a new Subevent model that go under the Activity model.

### Reason for Change

Completes CFD-102

### Type of change

### Testing

Works as described on Prisma


packages/db/prisma/schema.prisma
```diff
@@ -49,13 +49,23 @@ model Comment {
...
-  startTime       DateTime @db.Time()
+  day             DateTime   @db.Date
+  startTime       DateTime   @db.Time()
   durationMinutes Int
   leader          String
   location        String
+  subevents       Subevent[]
+}
+
+model Subevent {
```
BakerWJ, line 62: should this be called subactivity instead?

packages/db/prisma/schema.prisma
```diff
@@ -49,13 +49,23 @@ model Comment {
...
+  id         Int      @id @default(autoincrement())
+  name       String
+  startTime  DateTime @db.Time()
+  endTime    DateTime @db.Time()
```
BakerWJ, line 66: should this be duration 🤔

packages/db/prisma/schema.prisma
```diff
@@ -49,13 +49,23 @@ model Comment {
...
+  startTime       DateTime   @db.Time()
   durationMinutes Int
   leader          String
   location        String
+  subevents       Subevent[]
+}
+
+model Subevent {
+  id         Int      @id @default(autoincrement())
+  name       String
```
BakerWJ, line 64: i think we also want to add location to this model

BakerWJ: should be optional

BakerWJ (changes requested):

BakerWJ (approved):

PR #369, opened 2024-02-17
CFD-101: create user profile component

RyanL123:
### Reason for Change

Completes CFD-101

### Type of change

### Testing

image


RyanL123:
For some reason I can't get TRPC working with Expo. Not sure if it's a WSL issue or not, I'll do some further investigation on my Mac without WSL. For now this will just be a UI only PR. We can have a follow-up PR for metadata later as to not block existing work.

BakerWJ (commented):
can you integrate this with the info from clerk for the profile picture, email, name?

BakerWJ (approved):

PR #401, opened 2024-03-16
[CFD-117] Create Discussion Likes Endpoints

1zhaohel:
### Description

Created endpoints for liking and unliking posts.

### Reason for Change

Completes CFD-117

### Type of change

### DIFF
--- a/packages/api/src/router/like.ts
+++ b/packages/api/src/router/like.ts
@@ -50,18 +50,6 @@ export const likeRouter = createTRPCRouter({
       }),
     )
     .mutation(async ({ input, ctx }) => {
-      const existingLike = await ctx.db.likes.findFirst({
-        where: {
-          postId: input.postId,
-          userId: ctx.userId,
-        },
-      });
-      if (!existingLike) {
-        throw new TRPCError({
-          code: "NOT_FOUND",
-          message: "Like not found",
-        });
-      }
       await ctx.db.likes.deleteMany({
         where: {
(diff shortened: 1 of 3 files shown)


packages/api/src/router/like.ts
BakerWJ, line 1: this router needs to be added to the main router at https://github.com/uoftblueprint/centre-for-dreams/blob/main/packages/api/src/root.ts

packages/api/src/router/like.ts
```diff
@@ -0,0 +1,34 @@
...
+        postId: z.number().nonnegative(),
+      }),
+    )
+    .mutation(async ({ input, ctx }) => {
+      await ctx.db.likes.create({
+        data: {
+          postId: input.postId,
+          userId: ctx.userId,
+        },
+      });
```
BakerWJ, line 18: what happens here if a like for this post already exists?

BakerWJ (changes requested):

packages/api/src/router/like.ts
```diff
@@ -0,0 +1,72 @@
...
+      if (!existingLike) {
+        throw new TRPCError({
+          code: "NOT_FOUND",
+          message: "Like not found",
+        });
+      }
```
BakerWJ, line 64: this check can be removed

BakerWJ (changes requested):

BakerWJ (approved):

PR #415, opened 2024-03-23
CFD131: Deny login if user not approved

RyanL123:
### Description

If the user's email is not marked `isApproved` in clerk, they will not be able to proceed to the login code page. There will be 2 separate error messages displayed below the input box for
- User email is in clerk, but not approved
- User email is not in clerk

### Reason for Change

This is to follow the design in figma and also prevent unapproved users from logging in

Completes CFD-131

### Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Testing

Source: trust me bro

### DIFF
--- a/apps/expo/src/app/signincode.tsx
+++ b/apps/expo/src/app/signincode.tsx
@@ -2,18 +2,22 @@ import { useState } from "react";
 import { Image, Text, View } from "react-native";
 import Toast from "react-native-root-toast";
 import { useRouter } from "expo-router";
-import { SignedOut, useSignIn } from "@clerk/clerk-expo";
+import { useAuth, useSignIn } from "@clerk/clerk-expo";
 
 import FilledButton from "~/components/FilledButtons";
 import { TextInput } from "~/components/inputs";
+import { api } from "~/utils/api";
 import { getErrorMessage } from "~/utils/errorUtils";
 import Logo from "../../assets/logo.png";
 
 const SignInScreen = () => {
   const router = useRouter();
   const { signIn, setActive, isLoaded } = useSignIn();
-
+  const { signOut } = useAuth();
   const [code, setCode] = useState("");
+  const [approvalError, setApprovalError] = useState("");
+  const createUserWithClerkIdIfNotExists =
+    api.user.createUserWithClerkIdIfNotExists.useMutation();
 
   /**
    * Attempts to log the user in with the one-time passcode provided. The email used is the one provided in the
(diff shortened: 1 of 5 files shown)


BakerWJ:
how is a new user supposed to register if they are blocked from submitting if they don't have an account?

RyanL123:
@BakerWJ Like we discussed offline, I thought users are manually inserted/approved via the admin dashboard? Another option is to insert the user into our database but still deny login, and they'd have to manually be approved via the admin dashboard before being able to log in

packages/api/src/router/user.ts
```diff
@@ -15,6 +20,29 @@ export const userRouter = createTRPCRouter({
...
+   * Checks if the user with the given email is approved in clerk.
+   * Since we want this to be callable when the user is not authenticated, it must
+   * remain public. Thus, we should not return sensitive data and only include the
+   * bare minimum in the return value.
+   * @returns an object of the form {userExists: bool, isApproved: bool}
+   */
+  isEmailApproved: publicProcedure
+    .input(z.string())
+    .query(async ({ input }) => {
+      const users = await clerkClient.users.getUserList({
```
BakerWJ, line 33: use getUser here instead

BakerWJ (commented):
I'm not really a fan of how this is currently implemented -- I feel like we could have a screen after the user is logged in that shows that they don't have permissions instead. This would also allow anyone to figure out the enrolled user emails which isn't great

RyanL123:
@BakerWJ I was following the existing design on Figma, but I'm open to new implementations

>  I feel like we could have a screen after the user is logged in that shows that they don't have permissions instead.
Are you suggesting a new view after the logging in with the email code? Or is this something shown on the account page?

Alternatively we can also have protected procedures check that user is approved

packages/api/src/router/user.ts
RyanL123: getUser only has filter by `userId`, not the email

BakerWJ: 🤔 instead of storing the isApproved stuff in clerk, can we just make it so that only approved users exist in the db. Then you only need to check the db here?

RyanL123: So we'd need another table in the DB storing approved emails? 

Maybe `isApproved` should be a field in our DB instead of Clerk, and we insert everyone into the DB, but default their approval to `false`. The admin dashboard can set the approval status within our db (so clerk would store no metadata)

BakerWJ: I don't think we would need another table -- but yea adding the isApproved into our dashboard instead. I think that would work well and should be an easy change

apps/expo/src/app/signincode.tsx
```diff
@@ -24,16 +28,35 @@ const SignInScreen = () => {
...
       });
-      await setActive({ session: completeSignIn.createdSessionId });
-      router.replace("/account");
+
+      // If we have the id, that means the user successfully
+      // authenticated using Clerk. However we shouldn't update the
+      // active session just yet in case they're not approved
+      if (completeSignIn.id) {
+        const isApproved = await createUserWithClerkIdIfNotExists.mutateAsync(
+          completeSignIn.id,
```
RyanL123, line 47: This actually isn't the user id, I'll have to change up the implementaiton a bit

PR #416, opened 2024-03-24
CFD-133: Fix auth middleware participant unique key constraint

RyanL123:
### Description

This PR fixes the issue where paticipant IDs will be re-used when logging in with another account since it was hard coded to `1`. This PR will change to create a new unique participant object without hard coding the ID and linking it to the user.

### Reason for Change

`protectedProcedure` was unable to be used since it inserted the user into the database if they didn't exist on each run. This means that if the user logged in with a different account the insertion would fail because the participant id would clash.

Fixes CFD-63

Completes CFD-133

### Type of change

### Testing

Source: trust me bro

### DIFF
--- a/packages/api/src/trpc.ts
+++ b/packages/api/src/trpc.ts
@@ -131,16 +131,14 @@ const isAuthed = t.middleware(async ({ next, ctx }) => {
   // If the user isn't in our database yet, insert them
   // Also creating a new participant that's linked to them
   if (user == null) {
-    const participant = await ctx.db.participant.create({
-      data: {
-        name: "",
-      },
-    });
-
     user = await ctx.db.user.create({
       data: {
         clerkId: clerkId,
-        participantId: participant.id,
+        participant: {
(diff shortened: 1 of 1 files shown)


packages/api/src/trpc.ts
```diff
@@ -117,32 +117,33 @@ export const publicProcedure = t.procedure;
...
+        name: "",
+      },
+    });
+
+    user = await ctx.db.user.create({
+      data: {
+        clerkId: clerkId,
+        participantId: participant.id,
+      },
+    });
```
BakerWJ, line 145:
```suggestion
    user = await ctx.db.user.create({
      data: {
        clerkId: clerkId,
        participant: {
          create: {
            name: ""
          }
        }
      },
    });
```

BakerWJ: simplify to one query

BakerWJ (approved):

PR #417, opened 2024-03-28
CFD-130: persist notification settings in database

RyanL123:
### Reason for Change

Completes CFD-130

### Type of change

### Testing

After clicking the toggle buttons, the new states are persisted
image

### DIFF
--- a/apps/expo/src/app/account.tsx
+++ b/apps/expo/src/app/account.tsx
@@ -1,8 +1,10 @@
-import React, { useState } from "react";
+import React from "react";
 import { Image, Switch, Text, View } from "react-native";
 import { SafeAreaView } from "react-native-safe-area-context";
 import { Stack } from "expo-router";
 import { useAuth, useUser } from "@clerk/clerk-expo";
+import { useQueryClient } from "@tanstack/react-query";
+import { getQueryKey } from "@trpc/react-query";
 
 import FilledButton from "~/components/FilledButtons";
 import { api } from "~/utils/api";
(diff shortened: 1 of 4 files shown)


apps/expo/src/app/account.tsx
BakerWJ, line 1: this implementation seems a bit weird -- I'll make some changes on my own branch that I think simplify this a bit

apps/expo/src/app/account.tsx
```diff
@@ -69,10 +74,48 @@ const NotificationContainer: React.FC<NotificationContainerProps> = ({
...
+  const [notifications, setNotifications] = useState({
+    [NotificationType.ANNOUCEMENTS]:
+      currentNotificationSettings?.notificationOnAnnoucements ?? false,
+    [NotificationType.COMMENTS_ON_POSTS]:
+      currentNotificationSettings?.notificationOnPostComments ?? false,
+    [NotificationType.LIKE_ON_POSTS]:
+      currentNotificationSettings?.notificationOnPostLikes ?? false,
+    [NotificationType.SCHEDULE_UPDATES]:
+      currentNotificationSettings?.notificationOnScheduleUpdates ?? false,
+  });
```
BakerWJ, line 90: is there a reason for this to be stored in local state if they are already in the notification settings hook, and are updated after a mutation?

apps/expo/src/app/account.tsx
RyanL123: Just tested by rebasing on top of your change for invalidating queries. It works, but there's a delay on the visual change of the toggle button (presumable because the query is being re-fetched). I think we should just keep this implementation by temporarily holding local state to allow the frontend to render the change immediately.

BakerWJ: https://tanstack.com/query/v4/docs/framework/react/guides/optimistic-updates you can do an optimistic update while it is being refetched, we just want to make sure that what is displayed is always the same as server state

BakerWJ (approved):

PR #424, opened 2024-03-30
[CFD-128] Integrate createpost page with trpc

helenaglow:
### Description

Integrated post page with TRPC

### Reason for Change

Completes CFD-128

### Type of change

### Testing

https://github.com/uoftblueprint/centre-for-dreams/assets/113125436/624532d4-1a1f-4ab0-a99c-9bff33b738c4

### DIFF
--- a/apps/expo/src/app/createpost.tsx
+++ b/apps/expo/src/app/createpost.tsx
@@ -11,12 +11,28 @@ import { Stack, useRouter } from "expo-router";
 
 import FilledButton from "~/components/FilledButtons";
 import OutlinedButton from "~/components/OutlinedButtons";
+import { api } from "~/utils/api";
 import LeftArrow from "../../assets/arrow-left.svg";
 
 function CreatePost() {
   const [post, setPost] = useState("");
   const [images, setImages] = useState<string[]>([]);
   const { back } = useRouter();
+
+  const { mutate: createDiscussion } =
+    api.discussion.createDiscussion.useMutation({
+      onSuccess: () => {
+        back();
+      },
+    });
+
+  const createPost = () => {
+    if (post != "") {
+      createDiscussion({
+        contents: post,
+      });
+    }
+  };
 
   const clearState = () => {
     setPost("");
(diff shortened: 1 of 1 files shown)


helenaglow:
Two things to note:
- There was no design for errors so I didnt include any (it just doesnt create post or leave the screen) -> I can add a red line of text underneath or some sort of indicator that they didnt fill out the required information.
- Also I dont exactly know how create post will be used, but forum data values dont update when you go back to the screen. You have to refresh the page. -> I can fix that as well just let me know. I dont want to overstep

apps/expo/src/app/createpost.tsx
```diff
@@ -112,14 +128,12 @@ function CreatePost() {
...
-                back();
+                createPost();
               }}
             >
               <FilledButton
                 onClick={() => {
-                  console.log("Created post");
-                  back();
+                  createPost();
                 }}
```
BakerWJ, line 137: should the createPost be on both of these?

helenaglow: One of these doesn't work so it doesn't really matter. I just added for consistency

BakerWJ: can you remove the one that doesn't work then -- I don't think that a touchable opacity is needed over this button, so it can probably be changed to a view

helenaglow: It is needed cause the whole view is touchable without feedback to close the keyboard (the discussion is multiline so the return makes a new line, not close the keyboard). The touchable without feedback then makes all the buttons not work, unless wrapped with a touchable and put the action there. 🤔

helenaglow: Unless you have other ideas 🙏

BakerWJ: the filled button is still a touchable opacity though? why does it need to be wrapped in another one?

helenaglow: It just doesn't work for some mysterious reason. I can remove it and screen record when I get back home. 

I've tried to do a bunch of research but this seems to be the widely used solution.

helenaglow: https://github.com/uoftblueprint/centre-for-dreams/assets/113125436/77c756df-29c0-44a0-91eb-f0586ceb497d


Ignore the sizing, but I keep pressing the button and nothing happens when you remove touchable opacity.

helenaglow: Like theoretically there has to be a better way cause this is so inefficient and jank

BakerWJ (approved):

helenaglow:
I guess just f it we ball

PR #425, opened 2024-03-31
[CFD-109] Implement push notifications for announcements

jasonfyw:
### Description
Sends notifications to all devices with a valid pushToken when announcement is created. Also properly registers a device's pushToken when logged in.

### Reason for Change

Completes CFD-109

### Type of change

### Testing

Creating an announcement (using API panel) from a user other than the user shown on screen creates an announcement and sends a notification to all users other than the one creating it

https://github.com/uoftblueprint/centre-for-dreams/assets/40612523/846c10b1-3cfb-4a63-9bdc-5a469379c3f3

### DIFF
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -1,9 +1,11 @@
 import type { inferRouterInputs, inferRouterOutputs } from "@trpc/server";
 
+import { appRouter } from "./src/root";
 import type { AppRouter } from "./src/root";
+import { createCallerFactory } from "./src/trpc";
 
 export { appRouter, type AppRouter } from "./src/root";
-export { createTRPCContext } from "./src/trpc";
+export { createTRPCContext, createCallerFactory } from "./src/trpc";
 
 /**
  * Inference helpers for input types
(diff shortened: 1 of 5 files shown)


packages/api/index.ts
```diff
@@ -16,3 +18,5 @@ export type RouterInputs = inferRouterInputs<AppRouter>;
  * @example type HelloOutput = RouterOutputs['example']['hello']
  **/
 export type RouterOutputs = inferRouterOutputs<AppRouter>;
+
+export const createCaller = createCallerFactory(appRouter);
```
BakerWJ, line 22: remove code related to caller and just extract notfiication logic into its own function

packages/api/src/router/announcement.ts
```diff
@@ -2,8 +2,26 @@ import { Prisma } from "@prisma/client";
...
+) => {
+  const caller = createCaller(ctx);
+  await caller.notification.push({
+    pushTokens,
+    title,
+    body,
+    subtitle,
+  });
+};
+
```
BakerWJ, line 24: same as above comment -- just pull out notification stuff into its own function

packages/api/src/router/announcement.ts
```diff
@@ -31,18 +49,30 @@ export const announcementRouter = createTRPCRouter({
       z.object({
         title: z.string().trim().min(1).max(300),
         contents: z.string().min(1),
+        sendNotification: z.boolean().optional().default(true),
```
BakerWJ, line 52: let's not let this be optional and don't include a default

BakerWJ (changes requested):

PR #501, opened 2024-11-14
[CFD-154] Update Event Page

Dani3lx:
### Description
Implemented the the event update page, enabling user to be able to update all event fields (except for subactivities).

### Reason for Change
Completes [CFD-154](https://linear.app/uoftblueprint/issue/CFD-154/create-page-to-update-events)

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Other

### Testing

image
image
image


apps/nextjs/src/pages/eventupdate/EventUpdatePage.module.css
```diff
@@ -0,0 +1,3 @@
+.form_input {
```
sarinali, line 1: A minor nitpick on this, I would put this in the `apps/nextjs/src/styles` folder, do the import like `import styles from "../../styles/eventupdate.module.css";` instead for consistent logical group. Thoughts?

sarinali (commented):
LMK what you think about my comment. Another thing: I apologize for this, but the tickets say event however I think maybe we can pick some consistency, do you prefer to refer to an activity as an event? Or should we rename instances of event here to activity?

sarinali (approved):
LGTM

PR #532, opened 2025-02-02
Adventurer e/cfd 199 create post popup

Adventurer-E:
### Reason for Change

Fixes CFD-

Completes CFD-199

Contributes to CFD-

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Other

### Testing

- [ ] Test A
- [ ] Test B
Screenshot 2025-02-02 at 4 10 28 PM


Adventurer-E:
Works for small images, but not for large images.
Do we stop here, or shall we develop robust pipeline for large images?

sarinali (commented):
see comments on slack

apps/nextjs/src/env.mjs
```diff
@@ -10,6 +10,9 @@ export const env = createEnv({
     NODE_ENV: z.enum(["development", "test", "production"]),
     DATABASE_URL: z.string().url(),
     CLERK_SECRET_KEY: z.string(),
+    AWS_ACCESS_KEY: z.string(),
```
sarinali, line 13: this makes it required to have when launch development server, it will exit people's servers that don't have these keys. i would make this optional (`.optional()`)? or you can provide instructions to the team to put a dummy value here. I also do not have the keys so I cannot test the functionality of the upload

apps/nextjs/src/components/navbar.tsx
```diff
@@ -1,33 +1,53 @@
...
-        <button className="m-2 rounded-[24px] border border-[#2E4D90] p-2">
-          <a href="/absences/">Absentees</a>
-        </button>
-        <button className="m-2 rounded-[24px] border border-[#2E4D90] p-2">
-          <a href="/activities/">Calendar</a>
-        </button>
-        <button className="m-2 rounded-[24px] border border-[#2E4D90] p-2">
-          <a href="/inviteuser/">Invite New User</a>
+        <button
+          className="m-2 rounded-[24px] border border-[#2E4D90] p-2"
```
sarinali, line 45: I know this is temporary for viewing the functionality but you can remove this, revert this entire file if its easier

apps/nextjs/src/components/createpost.tsx
```diff
@@ -0,0 +1,370 @@
...
+                className={styles.profilePicture}
+                width={80}
+                height={80}
+                priority // Ensures the profile image is prioritized for loading
+              />
+            </div>
+            <CardTitle className="font-inter text-[18px] font-bold leading-[21.78px] tracking-[0%]">
+              User Name
+            </CardTitle>
+            {/* <CardDescription>Deploy your new project in one-click.</CardDescription> */}
```
sarinali, line 255: delete this line

apps/nextjs/src/components/createpost.tsx
```diff
@@ -0,0 +1,370 @@
...
+  const onSubmit = async (data: FormData) => {
+    if (data.title === "") {
+      console.error("Title is required.");
+      return;
+    }
+
+    // setUploading(true);
+    let uploadedImages: string[] = [];
+
+    try {
```
sarinali, line 82: I think you can refactor this logic into a separate class cuz your logic for checking the uploaded images and then making a request kind of lives outside this create post component

apps/nextjs/src/pages/api/upload_to_s3.ts
```diff
@@ -0,0 +1,59 @@
+import type { NextApiRequest, NextApiResponse } from "next";
```
sarinali, line 1: is this file meant to be created inside of the `pages` folder? i think a fitting place would be a new folder called `lib`. Unless you have justification otherwise

apps/nextjs/src/components/createpost.tsx
```diff
@@ -0,0 +1,370 @@
...
+      <form onSubmit={handleSubmit(onSubmit)}>
+        <Card
+          className="mih-[519px] w-[824px]"
+          style={{
+            position: "fixed",
+            top: "10%",
+            left: "20%",
+          }}
+        >
+          <button
```
sarinali, line 221: i think there is a button component from shadcn, do you mind using that instead?

apps/nextjs/src/components/createpost.tsx
```diff
@@ -0,0 +1,370 @@
...
+                      });
+                      return `data:image/png;base64,${btoa(binary)}`;
+                    };
+
+                    const base64String = uint8ArrayToBase64(image);
+                    return (
+                      <CarouselItem key={index}>
+                        <div className="flex h-full items-center justify-center p-1">
+                          <Card>
+                            {/* <CardContent className="flex items-center justify-center p-6"> */}
```
sarinali, line 307: delete

apps/nextjs/src/components/createpost.tsx
```diff
@@ -0,0 +1,370 @@
...
+      }
+    } catch (error) {
+      console.error("Error picking image:", error);
+    }
+  };
+
+  return ReactDOM.createPortal(
+    <>
+      <div
+        className="z-999 fixed inset-0 bg-black bg-opacity-50"
```
sarinali, line 202: i would make this scrollable if possible? 
image

bottom is cut off

apps/nextjs/src/components/createpost.tsx
```diff
@@ -0,0 +1,370 @@
...
+                }
+                alt="Profile"
+                className={styles.profilePicture}
+                width={80}
+                height={80}
+                priority // Ensures the profile image is prioritized for loading
+              />
+            </div>
+            <CardTitle className="font-inter text-[18px] font-bold leading-[21.78px] tracking-[0%]">
+              User Name
```
sarinali, line 253: can you pull the user's actual name? (userid is fine for now)

sarinali (commented):
really good work!! few things to address i think

apps/nextjs/src/env.mjs
sarinali: you can send them to me on slack :)

apps/nextjs/src/pages/api/upload_to_s3.ts
Adventurer-E: I prefer to keep it in "api", because it's an api router and i can easily use fetch("/api/upload_to_s3", ...) to fetch it. Otherwise, even if I moved it to lib, I still need to create a file in api folder to refer to this file.

apps/nextjs/src/components/createpost.tsx
Adventurer-E: I actually don't really understand. And I think it makes programming easier to let it live inside this file, because the front-end component (image displaying) also relies on the uploadedImages array. Moreover, the image array for each created post is separate, so I would prefer to let it stay here.

sarinali: Like what i mean from lines 82 to 139, is not dependant on the actual UI. You can put this into another fucntion, outside of this file for better modularity. What if I need to reuse this function somewhere? Will you copy this exact implementation again? What if there is another web page that is going to require submitting photos as well?

Again, on lines 142 to 195 of this file, the logic of this method is encoding it in base64 and setting your hook. you can create a function taht does the base64 conversion, and then return something and set that as your ImagesTemp. This is much cleaner and if someone needs to do another base64 conversion, they can reuse your method.

apps/nextjs/src/components/navbar.tsx
Adventurer-E: I'll temporary leave this temporary button (commented) instead of deleting it so that in the future people can easily refer to how to route to the createpost window.

PR #551, opened 2025-03-25
UTC Fix 1 WEB

sarinali:
Main goal of this PR is to consolidate UTC and local time storage. 

Displays on web are updated now to use the following rules:

- All time in the database should be in UTC. 
- Time should be converted on client side
    - All time coming into the server is in UTC
    - All time coming out of the server is in UTC
    - All time received from the API is converted on client side

Some very funny things I'd like to note:
- did you know that daylight savings officially began in 1970 for JavaScript? I know this because my times weren't converting correctly on client side because of the fact that daylight savings didnt start until after 1970. So then we had to manually fix the date to be after 1970 in order for the actual time conversion to work. FML... 

DEMO:

https://github.com/user-attachments/assets/f86f661c-bbce-47bb-9765-300dcca9d0cd


3milyfz (approved):
she cooked

